PR #1, opened 2017-06-15
change to use objects instead of strings

kentcdodds:
Hi there! I saw this and love it. Naturally I wanted to see whether it works with glamorous and found that it _almost_ does except it returns strings rather than objects. Turns out that styled-components can accept objects as well as strings! This is what `polished` does and you can use `polished` with styled-components or any other CSS-in-JS library.

Here's a start to making it work with objects. It's a work in progress. I've just updated the `fontSize` util. I've published these changes as `@kentcdodds/styled-system` and verified that it works with both glamorous and styled-components [here](https://codesandbox.io/s/xk32O0PXl).

If you're ok with this kind of change, I will be happy to update the pull request to apply to the other utils 👍


jxnblk:
Generally I'm all 👍 on this.

But, wait, what!? does styled-components support passing plain JS objects in?? I don't know where I missed that part of the docs...

kentcdodds:
Sweet! I'll finish the PR then!

README.md
```diff
@@ -128,8 +141,17 @@ All props accept arrays as values for mobile-first responsive styles.
 
 styled-system includes a higher order component to add style props to any component.
 
+**With `styled-components`**
+
+```jsx
+import hoc from 'styled-system/dist/styled-components/hoc'
```
kentcdodds: Here's a breaking change. Let me know if you'd prefer to go about this differently...

docs/Example.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
 import { ThemeProvider } from 'styled-components'
-import Box from '../src/Box'
+import Box from '../src/styled-components/Box'
```
kentcdodds: `Box` was never documented, so this is technically non-breaking?

test.js
```diff
@@ -1,15 +1,10 @@
 import test from 'ava'
-import {
-  space,
-  width,
-  fontSize,
-  util
-} from './src'
+import {space, width, fontSize, util} from './src'
 
 const theme = {
```
kentcdodds: Sorry, I was making tons of changes and formatted this with prettier. If you'd like me to revert stylistic changes, I can do that.

README.md
jxnblk: Why break this? An attempt to make it not dependent on styled components?

kentcdodds: Yeah, that was the intent. If people using other CSS-in-JS libs (glamorous, aphrodite, glamor, jss, etc.) they'll not want to load `styled-components` to use `styled-system`.

kentcdodds: In fact, most of this lib works without React at all as well, so I could see it being useful even beyond React too.

jxnblk: Yeah definitely see your point. Although it'd still install deps, what about a separate module in the root for styled-components and glamorous HOCs? e.g.:

```js
import hoc from 'styled-system/styled-components'
// or 
import hoc from 'styled-system/glamorous'
```

docs/Example.js
jxnblk: Yeah, I literally threw this together on my couch last night, so not sure that Box is needed here. The intent is to use this lib in grid-styled, etc

test.js
jxnblk: As long as you didn't add semicolons ;)

README.md
kentcdodds: That sounds fine!

> Although it'd still install deps

Actually `styled-system` doesn't currently have `styled-components` listed in `dependencies`. And I think that's fine, especially with this change :+1:

kentcdodds: I'll go ahead and make this update :+1:

test.js
kentcdodds: 🎉

README.md
jxnblk: 👍 fair warning I might change my mind later and rip out any library specific stuff

kentcdodds:
I figured out how the relab stuff worked 😅 and updated the examples for the new structure :+1:

README.md
kentcdodds: I'm totally cool with that :) :+1:

mxstbr:
> does styled-components support passing plain JS objects in?? I don't know where I missed that part of the docs...

Don't worry you didn't miss anything, it's not documented because we don't want people to use styled-components that way 😆  we added it to make `polished` work for all CSS in JS libraries, not just return strings and only support styled-components. :heart:

kentcdodds:
> it's not documented because we don't want people to use styled-components that way 😆

Maybe I shouldn't have [tweeted about it](https://twitter.com/kentcdodds/status/875435285307244544) then 😅  Or I should have been more clear that you _don't_ want to do that. (though, I don't know why anyone would want to!)

jxnblk:
@mxstbr is there any perf dent or disadvantage to using that then? I feel like the docs mention something about inline styles but didn't realize the stylis part could handle that?

README.md
```diff
@@ -1,6 +1,6 @@
 # styled-system
 
-Design system utilities for styled-components
+Design system utilities for CSS in JS (like styled-components, glamorous, etc.)
```
jxnblk: Also fair warning: I will probably change this copy (I hate the word *like*)

README.md
```diff
@@ -179,6 +201,7 @@ styled-system can be configured with styled-components'
 
 ```jsx
 import { ThemeProvider } from 'styled-components'
+// or import { ThemeProvider } from 'glamorous'
```
jxnblk: Don't go overboard with the marketing, pretty sure this would've worked with cxs months ago ;)

docs/Example.js
```diff
@@ -1,13 +1,12 @@
...
-import Box from '../src/Box'
-
+import { Box } from '../styled-components'
 
 const App = () => (
   <ThemeProvider theme={theme}>
     <Box p={[ 0, 1, 2, 3 ]}>
       <Box f={[ 5, 6, 7, 8 ]}>
-        Hello
+        Hello there
```
jxnblk: 👎

test.js
```diff
@@ -1,15 +1,10 @@
...
-  breakpoints: [ 32, 48, 64 ],
-  space: [ 0, 6, 12, 18, 24 ],
-  fontSizes: [ 12, 16, 18, 24, 36, 72 ]
+  breakpoints: [32, 48, 64],
+  space: [0, 6, 12, 18, 24],
+  fontSizes: [12, 16, 18, 24, 36, 72],
```
jxnblk: Okay, i don't like that prettier does this, but don't worry about it

jxnblk: This means I have to `i-return-esc` instead of `r-return`

jxnblk:
Alright, I'm done with my snarky comments, @kentcdodds - this looks fantastic. Will give it another review in the morning before merging. Thanks! 😎

README.md
kentcdodds: Just wanted to make it clear to people who are coming from glamorous. Happy to remove the styled-components import in favor of glamorous 😉 jkjk, I can remove this if you want.

docs/Example.js
kentcdodds: Whoops, didn't mean to commit that

test.js
kentcdodds: Huh? Sorry, not sure I follow...

kentcdodds:
Feel free to change whatever. I probably won't get to it until later.

README.md
jxnblk: Nah it's all good

mxstbr:
> is there any perf dent or disadvantage to using that then?

Maybe? I haven't benchmarked it or anything but essentially we have to parse twice, once objects -> strings and once strings -> final CSS. Unsure how slow that first part is though, honestly 🤷‍♂️

braposo:
Great job @kentcdodds, looking forward to seeing this released! Just a quick note though:

When using props like `fontSize` or `mt` the generated styles raise the `Warning: Unsupported style property margin-top. Did you mean marginTop?` warning. It still works but you get that error in the console. I think you need to change everything to pascalCase as `styled-components` will parse that anyway. (tests will break though!)

lucasmotta:
Nice one! And as a bonus point, it also works with [Fela](http://fela.js.org):
https://codesandbox.io/s/k5roZ2Krx

jxnblk:
@braposo I'm unable to reproduce locally, do you have an example anywhere?

jxnblk:
@kentcdodds this is great - going to merge it in. thanks again for all your help! 💯

kentcdodds:
Thank you! Can't wait for the release! I'll add it to the glamorous integrations page! 🎉

braposo:
@jxnblk it only raises the warning if you're using it with glamorous I think. @kentcdodds example in [here](https://codesandbox.io/s/xk32O0PXl) has that problem, just open the console and you'll see the warning for `font-size`.

braposo:
I've created another sandbox where that's more evident and it uses the latest version of `styled-system`. I'm happy to create a PR if you want.

braposo:
@jxnblk I just went ahead and created a PR for this: https://github.com/jxnblk/styled-system/pull/3

PR #3, opened 2017-06-18
Use camelCase for CSS properties

braposo:
As noted in #1, we get warnings in the console if using props like `fontSize` or `mt` because they're converted to `font-size` and `margin-top` properties. 
image

You can see that clearly [here](https://codesandbox.io/s/LgOMx85Zp) (just open the console and you'll see the warnings).

This converts everything to camelCase, which gets rid of the warnings and keeps compatibility with `styled-components` as expected. This can also be tested [here](https://codesandbox.io/s/lOvxm9jj).

(There's also a small typo fix for the README where it's importing the wrong module)

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ And array values are converted into [responsive values](#responsive-styles).
 ## space
 
 ```js
-import { fontSize } from 'styled-system'
+import { space } from 'styled-system'
 ```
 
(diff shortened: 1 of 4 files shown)


jxnblk:
Yeah, it seems like this is only an issue with glamorous from what I can tell. That said, camelCase makes more sense for object keys, and I much prefer plain JS objects over embedded CSS in template literals.

This looks great! Thanks for the help 😎

PR #5, opened 2017-06-20
make mq an export

dagda1:
mq is useful so it would be good as an export


jxnblk:
I wish I'd spent the time to give all these functions better names, but this PR makes sense to me – thanks!

Issue #9, opened 2017-06-28
Decouple sytled-system from glamorous and styled-components


josepot:
Hi and thanks for this library!

I've made a (quick and dirty) fork of [grid-styled](https://github.com/jxnblk/grid-styled) that I've named [grid-styletr](https://github.com/josepot/grid-styletr). It has the exact same API as grid-styled but it's built on top of [styletron-react](https://github.com/rtsao/styletron) rather than [styled-components](https://github.com/styled-components/styled-components). Therefore, my fork also depends on styled-system.

I'm using grid-styletr in a pretty big project and so far it's working great. The only "problem" that I have is that I'm getting warnings about an unmet peer dependency:

```
warning "glamorous@3.23.4" has unmet peer dependency "glamor@>=2"
```
That's because styled-system [has a dependency on glamorous](https://github.com/jxnblk/styled-system/blob/master/package.json#L39) and glamorous [has a peer-dependency on glamor](https://github.com/paypal/glamorous/blob/master/package.json#L40).

Styled-system is a great library. The fact that it can be used from glamour, styled-components, styletron and probably many other css-in-js libraries shows that styled-system has a great API. That's why I think that it should be completely decoupled from any css-in-js library.

I think that the ["Box" and the hoc](https://github.com/jxnblk/styled-system/blob/18420cd9e9a1dda646eb99e59d158dcd97adc749/src/styled-components/index.js) of both glamorous and styled-components don't belong to styled-system. Those are nice helpers that should be implemented from outside of styled-system, so that styled-system can stay completely decoupled from the different css-in-js libraries that can leverage it.

### UPDATE:

I was working on a PR for removing these dependencies and then I realized what the issue is: those helpers are only used from inside the "docs" (makes sense). So I guess that the solution is to have those dependencies as part of the dev-dependencies and the helpers declared inside the "docs" folder (no need for those helpers to be part of "dist"). That's pretty much what #10 does. :smile:

PR #10, opened 2017-06-28
Decouple styled-system from concrete css-in-js libraries

josepot:
Addresses #9


jxnblk:
Thanks for this! I was actually considering removing all the library-specific stuff back when #1 was merged in – I agree that it doesn't make much sense to be part of this package. Will take a look and probably merge soonish

Issue #7, opened 2017-06-26
Responsive directional space doesn't work as expected


angusfretwell:
Another great library, thanks!

I discovered a small issue. Applying responsive directional margins and paddings don't seem to work right. It seems to only apply the responsive padding or margin for the last defined direction, for example:

```
<MyComponent pt={[1, 2]} pb={[3, 4]} />
```

The result is that the bottom padding is applied correctly on both breakpoints, but the top padding is only correct on the smaller breakpoint.

PR #11, opened 2017-06-29
[FIX]: Responsive directional space

josepot:
I was also affected by #7 so I decided to have a look at it.

The problem is that the [current `joinObj` method](https://github.com/jxnblk/styled-system/blob/e21b4212b14f557f9a10c1f95fab654ce63b2b73/src/util.js#L16) does a shallow merge, which screws up the "media" entries (because those are objects).

Thank you so much @angusfretwell for those tests, they've made the fix super easy. I'm including your commit in my PR, so I guess that if this PR gets merged you should also get credit for your contribution. I hope that's ok with you :wink:


angusfretwell:
Woohoo, thanks for taking the time to fix this. Glad the tests came in handy!

jxnblk:
Thanks, this looks great! I'll probably export `util.joinObj` and add tests for that as well since it's not currently covered

PR #22, opened 2017-08-09
Fix responsive-style key for multi value

b2whats:
https://github.com/b2whats/styled-system/blob/ef19a1601421e2845a8c21ccc1514863f38a367f/test.js#L570-L577

the component does not work correctly in the multiple values

### DIFF
--- a/src/responsive-style.js
+++ b/src/responsive-style.js
@@ -1,28 +1,36 @@
 const { is, idx, arr, num, px, breaks, dec, media, merge } = require('./util')
 
-module.exports = (key, prop, boolValue) => props => {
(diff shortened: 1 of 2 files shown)


jxnblk:
Thanks, this seems more like a "feature" than a "fix" and I'm not really sure of the value of this for any prop other than the one you have in the test... Leaving this open for discussion though

jxnblk:
Thanks for putting this together, but I think you could accomplish a similar thing by wrapping the `responsiveStyle` function, and I don't feel like I want to introduce more complexity into this library until it's refactored a bit.

PR #24, opened 2017-08-17
Support alternate units for media queries

ifyoumakeit:
Currently `styled-system` only supports `em`s for media queries. This PR will allow users to use any unit when specifying theme breakpoints. 

- Add `em` function to append `em` to number
- Update README to show usage
- Test `em` function
- Test `px` media query w/ `util.break`

Looking to bring styled-system into our design system ~~and this is the only blocker~~. 

---
Edited: Not super important because we can also convert breakpoint units to ems for `styled-system`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,6 +217,7 @@ const Component = styled(BaseComponent)([],
 styled-system uses a mobile-first responsive approach,
 where any value set works from that breakpoint and wider.
 The default set of breakpoints aims to cover a wide range of devices from mobile to desktop.
+Breakpoints default to `em` but can be overridden by passing strings with unit appended.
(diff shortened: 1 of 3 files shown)


b2whats:
can be done simply
```js
const mq = n => `@media screen and (min-width: ${n})`
```

ifyoumakeit:
@b2whats that would be a breaking change, since it previously supported the automatic addition of `em` with integers. This also follows how `px` is added to font-sizes and widths.

Users would get `@media screen and (min-width: 30)` which wouldn't work.

jxnblk:
Thanks for this and thanks for making sure to not introduce a breaking change! I've had the question about why breakpoints are specified in px units, but I think this solution is a nice, non-breaking approach to support other units

PR #26, opened 2017-08-23
Clarify arbitrary array lengths for theme configs

colepeters:
Adds a note to clarify that arrays passed to `breakpoints`, `space`, and `fontSizes` theme props can be of arbitrary lengths. 🤙


jxnblk:
Awesome, thanks!

PR #31, opened 2017-09-09
Allow strings in spacing scale

### LINKED ISSUE #30: Support string values in the spacing scale
It would be great if I could put em strings in the spacing scale. For example:

```javascript
const a = space({
  theme: { space: ['1em', '2em'] },
  m: 1
});
console.log(a); // { margin: NaN }
```

Right now they're assumed to be numbers so when the value is multiplied by 1 or -1 to follow the rule
> Negative values can be used for negative margins

we get NaN.

Maybe we could check the type of the index in the array before we multiply it by a number?

smith-kyle:
This will allow people who are using non px units to pass in string values in the spacing scale. Like:
```javascript
const a = space({
  theme: { space: ['1em', '2em'] },
  m: 1
});
console.log(a); // { margin: '2em' }
```

Something I'm unsure about with this change is the case where a user has string values in the spacing scale and passes in a negative value for a prop that uses the spacing scale. So for example:
 ```javascript
const a = space({
  theme: { space: ['1em', '2em'] },
  m: -1 // <-- Normally would translate to themes.space[1] * -1
});
console.log(a); // { margin: '2em' }
```
Now that the value is a string we can't safely make it negative without doing some parsing

Closes #30

### DIFF
--- a/src/space.js
+++ b/src/space.js
@@ -38,9 +38,18 @@ module.exports = props => {
   }).reduce(merge, {})
 }
 
-const mx = scale => n => num(n)
-  ? px((scale[Math.abs(n)] || Math.abs(n)) * (neg(n) ? -1 : 1))
-  : n
+const mx = scale => n => {
+  if (!num(n)) {
+    return n
+  }
+
+  const value = scale[Math.abs(n)] || Math.abs(n)
+  if (!num(value)) {
+    return value
+  }
+
+  return px(value * (neg(n) ? -1 : 1));
+}
 
 const getProperties = key => {
   const [ a, b ] = key.split('')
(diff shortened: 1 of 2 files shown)


smith-kyle:
Does this look ok? Or are you a fan of the value parsing so negative indices turn into `themes.space[i] * -i`

jxnblk:
Thanks for putting this together! Yeah, I think if a user sets string values for the spacing scale, then it's fine to not support the negative margin case. It wouldn't be a breaking change, but could be a feature to add later

colepeters:
Aaaaamazing, thanks for adding this @smith-kyle! I was going to be looking into trying to implement this myself in the next couple weeks, but noticed this addition today while reviewing the 1.0 release. Much appreciated.

PR #37, opened 2017-09-29
Preset style function props

ifyoumakeit:
- Create `preset` function
- Export from library
- Add documentation/test

I ran into this issue when trying to override a button's base background color. I initially reordered my style functions to the end of the template literal, but this resulted in two properties with the same name being set in the style tag with the later property in the cascade winning.

image

This PR adds a short-hand function `preset` that takes a style function and a fallback object as arguments and returns a new style function with defaults set.

[CodeSandbox example w/ longhand & shorthand version](https://codesandbox.io/s/0xl49my5nv)

```jsx
// Double styles
styled.a`
  ${color}
  color: red;
`

// Current long-hand
styled.a`${props => color(Object.assign({}, { color: "green" }, props)}`
styled.a`${props => color({ color: "green", ...props })}`

// Short-hand version
styled.a`preset(color, { color: "green" })`
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -393,6 +393,35 @@ const Component = styled(BaseComponent)([],
 )
 ```
 
+## Preset Style Functions
+
+`preset` allows you to set fallback props for your style functions.
+Most CSS-in-JS libraries don't strip out duplicate properties, this
+requires that you ensure proper ordering of styles.  
+
+
+```jsx
+import React from 'react'
+import styled from "styled-components"
+import { fontSize, space, color, preset } from "styled-system"
+
+// Example of named preset.
+const presetFontSize = preset(fontSize, { f: [24, 36, 60, 72] })
+
+const Button = styled.button`
+  ${presetFontSize} 
+  ${preset(color, { color: "red" })} 
+  ${space}
+`;
+
+const App = props => (
+  <div>
+    <Button f={[36, 40, 44, 50]}>Classic red button</Button>
+    <Button color="blue">New blue button</Button>
+  </div>
+)
+```
+
 ---
 
 ## Low-level style functions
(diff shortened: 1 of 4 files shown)


ifyoumakeit:
This definitely falls under the "nicety" column, this utility could just be up to the user to define.

jxnblk:
Thanks! Yeah I’m a little confused as to how this is different/better than default props. Part of the idea with this library is to move any configurable/dynamic/themed styles into React’s props API and outside of CSS declarations

ifyoumakeit:
Makes sense, my initial thinking is that styled-system defaults should sit with the styled-system injection into the template literal.

I think I was spending a little too much time in styled-system land instead of React. Closing!

dzucconi:
While although we can indeed define it; this would be useful.

I might be missing some other obvious way to accomplish this but consider the use-case of a mixin:
```javascript
import { css } from 'styled-components';
import { fontSize } from 'styled-system';

const mixin = css`
  ${fontSize}
`;
```

How would one go about setting a default value here?

dzucconi:
Something I'm wondering about here is how you could handle setting presets while still supporting shorthand aliases.

Something to consider is that defaultProps won't work for a mixin. So if you do want to set presets inline like this, you have to always be sure to specify the aliased version of the prop when setting a default value because the proper name will always take precedence:

https://github.com/jxnblk/styled-system/blob/31687a4c014ece88e4dabbfcfdc006a66b36b43e/src/util.js#L71

```javascript
css`
  ${preset(fontSize, { fontSize: 5 })}
`

<Element fontSize={2} /> // => fontSize is 2
<Element f={2} /> // => fontSize is still 5

// vs

css`
  ${preset(fontSize, { f: 5 })}
`

<Element fontSize={2} /> // => fontSize is 2
<Element f={2} /> // => fontSize is 2
```

So I can see the value of some kind of built-in that takes this into consideration.

I might be missing some other obvious way to be setting values in situations like this?

PR #39, opened 2017-10-01
Fix border color

ferdinandsalis:
border color currently uses the style helper which does not use the `get` for retrieving nested values. Hence values like `borderColor="gray.4"` are ignored.

I have considered just extending the color helper but that would mean that `borderColor` would be responsive but since none of the other border helpers are responsive I kept it that way.

Feel free to discard or whatever …

🙏

### DIFF
--- a/src/style.js
+++ b/src/style.js
@@ -7,8 +7,7 @@ module.exports = ({
 }) => props => {
   const n = props[prop]
   if (!is(n)) return null
-  const scale = get(props, [ 'theme', key ].join('.'), {})
-  const val = scale[n] || n
+  const val = get(props, [ 'theme', key, n ].join('.'), n)
 
   return { [cssProperty || prop]: val }
 }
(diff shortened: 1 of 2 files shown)


jxnblk:
Thanks! I'm a little confused... style helper *should* be using the `get` accessor (https://github.com/jxnblk/styled-system/blob/master/src/style.js#L10) – if it's not, then that's a bug with the `style` function

ferdinandsalis:
Yes thats correct style is using it but not for the value — https://github.com/jxnblk/styled-system/blob/master/src/style.js#L11. The `get` on Line 10 is only for particular theme prop but not the value itself.

jxnblk:
Ah yeah I see now. I think the value there should be using the get utility if you want to PR that

ferdinandsalis:
Coming right up!

ferdinandsalis:
Done!

jxnblk:
Awesome thanks! 😎

PR #43, opened 2017-10-04
Fix cleanElement refs

oomathias:
`cleanElement` shouldn't be a stateless function, so we can pass refs through the tree.


jxnblk:
LGTM

PR #47, opened 2017-10-11
added babel-engine-plugin

jonstuebe:
Hey, I noticed that recently dot-prop updated and broke the build because of their lack of es6 support. I think this should solve the problem but don't have a good way of testing. Can you take a look?

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "ava": "^0.19.1",
     "babel-cli": "^6.24.1",
     "babel-core": "^6.25.0",
+    "babel-engine-plugin": "^0.1.2",
     "babel-loader": "^7.0.0",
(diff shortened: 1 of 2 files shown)


jxnblk:
dot-prop has been removed from the dependencies

oomathias:
@jonstuebe If you are using webpack and do not want to upgrade styled-system (which will resolve your issue).

You can do this hack in your resolver:
```js
  loaders: [
    {
      test: /\.js$/,
      loader: 'babel-loader',
      exclude: /node_modules\/(?!(dot-prop)\/).*/
    }
  ]
```

jonstuebe:
@oomathias & @jxnblk. Thx guys for the quick replies!

PR #49, opened 2017-10-12
Renamed prop-types.js to propTypes.js.

aVozgrin:
Bug with prop-types name described in [#51](https://github.com/jxnblk/grid-styled/issues/51)


oomathias:
This is not a bug. See https://github.com/jxnblk/grid-styled/issues/51#issuecomment-336222958

Excluding  the`src` folder from the npm package, is a way better option IMO.

jxnblk:
Closing this out in favor of #54 – I would recommend updating your config aliases to avoid potential issues like this in the future

PR #51, opened 2017-10-13
WIP: remove empty css blocks

oomathias:
Cleanup the generated css by removing all the empty blocks.

As you can see bellow, null doesn't generate useless blocks like false do. I created a lib on top of `styled-system` and the difference on a big app is huge.

```js
test('responsiveStyle allows array values', t => {
  const direction = responsiveStyle('flex-direction', 'direction')
  const a = direction({ direction: [ 'column', null, 'row' ] })
  const b = direction({ direction: [ 'column', false, 'row' ] })
  const result = 
  t.deepEqual(a, {
    'flex-direction': 'column',
    '@media screen and (min-width: 52em)': {
      'flex-direction': 'row',
    }
  })
  t.deepEqual(b, {
    'flex-direction': 'column',
    '@media screen and (min-width: 40em)': {
      'flex-direction': false
    },
    '@media screen and (min-width: 52em)': {
      'flex-direction': 'row',
    }
  })
})
```

I may have forgotten some parsers like border.

### DIFF
--- a/src/responsive-style.js
+++ b/src/responsive-style.js
@@ -22,11 +22,12 @@ module.exports = (...args) => props => {
   const sx = val => get(scale, '' + val, numberToPx ? px(val) : val)
 
   if (!Array.isArray(n)) {
-    return {
-      [cssProperty]: sx(
-        bool(boolValue)(n)
-      )
-    }
+    const val = sx(
+      bool(boolValue)(n)
+    )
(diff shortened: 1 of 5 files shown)


jxnblk:
Sorry for not following up on this. The code base has moved around a bit since this was opened, but feel free to open an issue if this is still something you'd like to contribute

PR #55, opened 2017-10-17
fix: added border color prop type

### LINKED ISSUE #45: Missing propTypes.borderColor
The `propTypes` export doesn't define borderColor, so the cleanElement HOC doesn't strip this prop and it's passed down to the styled element.

theutz:
This seemed like a pretty straightforward fix
Fixes #45

### DIFF
--- a/src/prop-types.js
+++ b/src/prop-types.js
@@ -92,6 +92,10 @@ const borderWidth = {
   borderLeft: PropTypes.bool
 }
 
+const borderColor = {
+  borderColor: PropTypes.string
+}
+
 const boxShadow = {
   boxShadow: PropTypes.string
 }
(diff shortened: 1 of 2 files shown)


theutz:
I also noticed that the PropTypes listed for box shadow do not align with the behavior exhibited, nor with the documentation. So I made a lil' fix!

src/prop-types.js
```diff
@@ -92,8 +92,15 @@ const borderWidth = {
   borderLeft: PropTypes.bool
 }
 
+const borderColor = {
```
jxnblk: Thanks for this! This will also need to be exported from the module

theutz: Good catch, @jxnblk. I took care of it.

jxnblk:
LGTM! Thanks again!

PR #57, opened 2017-10-19
rename prop-types.js -> src/proptypes.js

igoldny:


maxdeviant:
I don't think this is the correct way to handle this.

There's already a `proptypes` module on npm ([link](https://www.npmjs.com/package/proptypes)), and changing the filename is just going to mask the issue.

The root of the problem is that `module.resolve` is too inclusive, so that should be the thing that changes, not `styled-system`.

igoldny:
ok thnx

PR #64, opened 2017-11-22
Add support for tracking

johnotander:


src/letter-spacing.js
```diff
@@ -0,0 +1,7 @@
+const style = require('./style')
+
+module.exports = style({
+  prop: 'letterSpacing',
+  key: 'letterSpacings',
```
johnotander: Key sounds funny to me, went back and forth between "letterSpacings", "letterSpacing", and "tracking".

jxnblk: I'd prefer sticking to CSS naming conventions, even if they aren't aligned with typographic terminology – I think `letterSpacings` is fine

PR #65, opened 2017-11-22
Add maxWidth support

johnotander:


src/measure.js
```diff
@@ -0,0 +1,27 @@
...
+  const bp = breaks(props)
+
+  return n
+    .map(mx(scale))
+    .map(dec('width'))
+    .map(media(bp))
+    .reduce(merge, {})
+}
+
+const mx = scale => n => num(n) ? px(scale[n] || n) : n
```
johnotander: Just realized this should be `em` : )

src/measure.js
```diff
@@ -0,0 +1,27 @@
...
+module.exports = props => {
+  const n = props.measure
+
+  if(!is(n)) return null
+
+  const scale = get(props, 'theme.measures', measures)
+
+  if (!Array.isArray(n)) {
+    return {
+      width: mx(scale)(n)
```
jxnblk: I feel like this should use `max-width` and the utility should be called `maxWidth` – sticking to CSS naming conventions like the rest of (most of??) the library

src/constants.js
```diff
@@ -24,8 +24,15 @@ const fontSizes = [
   72
 ]
 
+const measures = [
+  24,
+  32,
+  48
```
jxnblk: In the next major version, I plan on changing the em-unit breakpoints to be pixel values to keep everything consistent. Would be great to handle this with pixels as well, but still allow em-units to be defined with a string

src/measure.js
johnotander: Yeah, `max-width` makes sense to me. I've updated my pr to use the css naming.

src/max-width.js
```diff
@@ -0,0 +1,27 @@
+const { get, is, px, num, breaks, dec, media, merge } = require('./util')
+const { maxWidths } = require('./constants')
```
jxnblk: I've been leaning towards fewer (potentially none) default values in styled-system – thinking it's better to work off of raw values, then let users set their own values in their theme – and I think max-widths are one of those things that probably vary widely from project to project

README.md
```diff
@@ -324,6 +325,14 @@ import { boxShadow } from 'styled-system'
 // props.theme.shadows[1]
 ```
 
+### maxWidth (responsive)
+
+```js
+import { measure } from 'styled-system'
```
jxnblk: Typo?

src/max-width.js
```diff
@@ -0,0 +1,27 @@
...
+const width = require('./width')
+
+module.exports = props => {
```
jxnblk: Might be missing something, but is there a reason not to use the `responsiveStyle` utility for this?

johnpolacek: That's what I do - https://github.com/johnpolacek/styled-system-html/blob/ecfc999347102adabe81dca43e7d5271d29ce66a/src/styles.js#L7 - works great :)

PR #67, opened 2017-11-28
Add peerDependency on react

Andarist:
Should probably work on older Reacts too, but I was not sure what the perfect range is.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,5 +63,8 @@
   },
   "dependencies": {
     "prop-types": "^15.6.0"
+  },
+  "peerDependencies": {
+    "react": "^0.14.x || ^15.x || ^16.x"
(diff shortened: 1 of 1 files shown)


jxnblk:
Thanks for this, but why add React as a peer dependency? This library should only be utility functions that take a `props` object and returns a `style` object

Andarist:
`cleanElement` uses `React` - https://github.com/jxnblk/styled-system/blob/208453adfb715527fdc313f72d026f3b5abc73a2/src/clean-element.js#L1

jxnblk:
Ugh, right. Sorry I forgot about that one – hopefully can get rid of that in the next version. Thanks for putting this together!

PR #68, opened 2017-11-28
Refactored to es modules

Andarist:
Benefits:
- tree-skakeability

Downsides:
- breaking change for people using `import system from 'styled-system'` - it needs to be `import * as system from 'styled-system'`


jxnblk:
Thanks! While I don't mind the idea of using ES modules, I don't want to do this in the library until the next major version. It also looks like your tests are failing, so closing this for now

Andarist:
> Thanks! While I don't mind the idea of using ES modules, I don't want to do this in the library until the next major version.

Are there any plans for the future major version right now? es modules can be added in non-breaking fashion.

> It also looks like your tests are failing, so closing this for now

I've pushed out the fix for tests.

PR #76, opened 2018-01-29
Add example fallback radius for theme function in README

ckeeney:
I assumed the fallback was just a second arg to the `theme` function, but I had to check the source to be sure.  This PR just includes the fallback in the example in the README.


jxnblk:
Awesome, thanks!

Issue #82, opened 2018-02-06
v2


jxnblk:
The following items need to be written up as issues and tagged with the v2 milestone.

- [x] convert to ES module syntax
- [x] remove idx
- [x] remove legacy responsiveStyle args parsing
- [x] consolidate style functions into single file
- [x] consolidate low-level utilities into single file
- [x] deprecate remove-props
- [x] use verbose names for flex properties
- [x] Numbers in `breakpoints` should convert to pixels
- [x] update default theme #91
- [x] add accessor argument to low-level utilities #87 #50
- [x] Ensure style prop names are unique #89 
- [x] Add alias option to low-level utilities #90
- [x] Remove `boolValue` option & boolean props
- [x] Normalize border utilities api #98
- [x] subpackages for monorepo
  - [x] clean-element
  - [x] clean-tag
  - [x] system-components
- [x] rename theme getter util #88
- [ ] update readme docs

**For consideration** these can be handled as separate non-breaking changes
- compose function for space & color
- props combinations utility
- add support for verbose margin and padding props
- [x] ~~add more docs w/ live examples~~ (not a breaking change and can be a follow-up task)

PR #83, opened 2018-02-06
Begin working on v2

johnotander:
#82


package.json
```diff
@@ -21,49 +17,10 @@
   "author": "Brent Jackson",
   "license": "MIT",
   "devDependencies": {
+    "@compositor/styleguide": "0.0.8",
```
jxnblk: I'm going to remove this dependency until it's public so that anyone can contribute

core/.babelrc
```diff
@@ -0,0 +1,10 @@
+{
+  "presets": [
+    "env",
+    "stage-0",
+    "react"
+  ],
+  "plugins": [
+    "transform-runtime"
```
jxnblk: I suspect this is only needed for docs with x0

core/src/index.js
```diff
@@ -0,0 +1,136 @@
...
+  // TODO: Directions option for styles
+  borderWidth,
+
+  // other
+  propTypes,
+  removeProps
+}
+
+// typography
+export const textAlign = responsiveStyle({
```
jxnblk: I think these could be in their own module, maybe called `styles`??

jxnblk:
Going to merge this to the `next` branch, but probably nuke the docs changes

PR #101, opened 2018-02-13
v2

### LINKED ISSUE #82: v2

jxnblk:
## What's new

- More utilities for CSS properties
- Uses verbose naming convention for `textAlign`, `alignItems`, `justifyContent`, `flexWrap`, and other CSS properties
- `style` and `responsiveStyle` utilities now accept a `getter` argument for converting values
- New `borders` utility that normalizes the API with other styled-system utilities
- Rewritten with ES module syntax
- DRYer code base for utilities like `fontSize`, `width`, and `color`
- The repo has been reorganized as a monorepo

### Breaking
- The `cleanElement` utility is now a separate package, removing the `react` dependency from the core library
- The `flexWrap` utility now expects a string value for its prop (previously a boolean)
- The `borderWidth` utility has been replaced with the `borders` utility 
- The `borderTop`, `borderRight`, `borderBottom`, and `borderLeft` props no longer accept boolean props, but follow the same convention as other utilities
- The default spacing scale has changed to `[ 0, 4, 8, 16, 32, 64, 128, 256, 512 ]` (this only affects usage without a custom `theme.space` provided
- Number values for breakpoints are now converted to px instead of ems to better match the rest of the library
- The `theme` utility has been renamed to `themeGet` (alternative names under consideration)
- Removes `removeProps` utility
- Removes `idx` utility
- Removes legacy `responsiveStyle` arguments API
- The `pseudoStyle` arguments API has changed to more closely align with `style` and `responsiveStyle`

**Shims for breaking changes**

To make migration easier, some breaking changes will continue to work as intended or similarly:
- `justifyContent` is aliased to the `justify` prop
- `textAlign` is aliased to the `align` prop, though this may cause unexpected behavior when used with the `alignItems` prop
- `alignItems` is aliased to the `align` prop, though this may cause unexpected behavior when used with the `textAlign` prop
- `flexWrap` is aliased to the `wrap` prop, but a string value of `wrap` or `nowrap` should be provided. (a shim to convert boolean `true` values to `wrap` has been added as well)
- `borderWidth` now works similarly to the new `borders` utility, which will continue to work for most use-cases
- The `themeGet` utility is aliased to a `theme` export

### New utilities

- fontFamily
- display
- minWidth
- height
- maxHeight
- minHeight
- size (width & height)
- alignContent
- order
- borders
- backgroundImage
- backgroundSize
- backgroundPosition
- zIndex
- top
- right
- bottom
- left

Fixes #82


Andarist:
I'd strongly encourage you to compile dist files with rollup to achieve a flat bundles (which should result in less code included in consumers' apps) and distribute both main and module entries (latter for treeshake-ability). If you want I can set up this tooling.

jxnblk:
@Andarist sorry that's not something I want to introduce to this library at the moment. A lot of the work here was to help simplify the code base for outside contributors, and I feel that adding more build setup sort of goes against that, especially for such a small utility library like this

JoshRosenstein:
@jxnblk - under utils you have propTypes with the key of '**responsivePropType**', and under responsiveStyle function you reference propTypes.responsive

Andarist:
> @Andarist sorry that's not something I want to introduce to this library at the moment. A lot of the work here was to help simplify the code base for outside contributors, and I feel that adding more build setup sort of goes against that, especially for such a small utility library like this

I don't feel **slightly** different build tooling would be a barrier for outside contributors.
1. most contributors are focused on the code and not the surrounding tooling - if we make it work, noone will even notice it. And it's not like you have no tooling at the moment
2. `rollup` is quite standard tool at the moment, used by numerous libraries and its config is straight-forward

As the author you are of course entitled to decide whatever you feel is right, I'd only want to raise this concern, because, for such a modular library as this, the improved tooling would give users real benefits and I can prepare demons backing that statement with numbers if you want.

JoshRosenstein:
@jxnblk - another suggestion: adding the getter along with pseudo:
apply a getter key just like numberToPx, Moved th variable outside of loop, apply the getValue function which allows the removal of the conditional PX conversion at the end.
````
export const pseudoStyle = (pseudoclass, prop) => (keys = {}) => props => {
    const style = props[prop || pseudoclass]
    const numberToPx = keys.numberToPx || {}
    const getters = keys.getter || {}
    const th = fallbackTheme(props)
    for (let key in style) {
        const toPx = numberToPx[key]
        if (!keys[key] && !toPx) continue
        const themeKey = [keys[key], style[key]].join('.')
        style[key] = getValue(get(th, themeKey, style[key]), getter, numberToPx)
    }
    return {
        ['&:' + pseudoclass]: style
    }
}

**Comment by jxnblk** [comment]
@JoshRosenstein that's a great idea, thanks! I should probably make the argument for `pseudoStyle` follow the other utilities as well

File: core/src/styles.js [hunk 1]
```diff
@@ -0,0 +1,354 @@
...
+    color: 'colors',
+    backgroundColor: 'colors',
+    borderColor: 'colors',
+    boxShadow: 'shadows'
+  }
+})
+
+// for backwards-compatibility
+// these will be removed in v3
+const __DEV__ = (process.env.NODE_ENV !== 'production')
```
Andarist: @jxnblk not every tool is able to inline constants, as it's used only once anyway I'd suggest to inline this check in the `getter` bellow

PR #106, opened 2018-02-16
Quick Fix on pseudoStyle new api

JoshRosenstein:
Never done a PR before, let me know if you prefer me to just shoot you a post on something i see or if I'm doing this correctly


jxnblk:
Good catch, thanks! This LGTM

PR #107, opened 2018-02-16
clean-tag: Fix bug preventing props from getting passed by exported HTML tags

shanecav:
This fixes a bug in `clean-tag` causing props passed to exported HTML tags (e.g. `tag.h1`) to not get passed on to children.

Here's a little demo showing the issue & solution: https://codesandbox.io/s/xp60z1v33o


jxnblk:
Awesome, thanks!

PR #118, opened 2018-02-21
system-components: use clean-tag

shanecav:
Removes the `Tag` component, and uses `clean-tag` instead.


jxnblk:
Awesome, thanks!

PR #119, opened 2018-02-21
RFC: system-components: include custom strings in blacklist

shanecav:
This change would allow you to include any string as an argument to `system`, and it would be added to the blacklist even if it doesn't exist as a `style` key in `styled-system`.

The benefit of this would be that you could use custom props in your system-components to control styling, without having them passed down to child components.

An example of this would be something like:

```js
import system from "system-components";

const Thing = system(
  {
    p: 4,
    bg: "blue",
    color: "white"
  },
  props => ({
    minHeight: props.tall ? "400px" : "200px"
  }),
  "tall"
);

render(<Thing tall />)
```

This could already be achieved using `clean-element`, but this way would be a little simpler, and wouldn't add an extra component to the tree.

Potential downside: Right now string arguments mean something really specific - this change would make the API a little less explicit. Users would have to be careful about their custom prop names, to avoid accidentally adding styled-system props. One possible way around this would be to require users to import the `styled-system` functions and add them as arguments directly (instead of using their string names), but that would be a breaking change.

Let me know what you think! If it seems like a good idea, I can add a test and update docs.


shanecav:
Yeah, in it's current state my PR is pretty half-baked, I'm just kind of thinking out loud trying to see if there could be a nice way to let ~~me~~ users filter out custom props.

One possible API would be something like:

```js
import system from "system-components";
import { borderRadius, fontSize } from "styled-system";

const Thing = system(
  {
    p: 4,
    bg: "blue",
    color: "white"
  },
  props => ({
    minHeight: props.tall ? "400px" : "200px"
  }),
  // Use styled-system style functions directly (their prop names
  // could be internally added to the blacklist via fn.propTypes)
  borderRadius,
  fontSize,
  // String arguments explicitly mean "just add this to the blacklist"
  "tall"
);

render(<Thing tall />)
```

That would fix the explicitness issue, and the implementation might even be a little simpler. Would be a breaking change, although it would be possible to support both the current API and this one at the same (gaining backwards compatibility at the cost of explicitness).

Again, just thinking out loud though - thanks for the feedback!

jxnblk:
Yeah, I wouldn’t want to remove the use of styled-system key strings, since that’s one of the main reasons for having this module.

Another thought: the first props argument could maybe have a `blacklist` prop, similar to the `is` prop...

shanecav:
Oh yeah, I like that idea a lot! I know you can already set a `blacklist` in that first `defaultProps` argument, but it currently overwrites the `blacklist` created internally if you do that. If it merged with it instead of overwriting it, that would be perfect.

I can update this PR if that's something you'd be interested in.

shanecav:
@jxnblk Not sure if you got notified or not, but I updated this PR to support setting a `blacklist` prop in the first argument.

PR #148, opened 2018-03-06
Remove the css grid doc section

johnotander:
It can be confusing to folks used to float/flex
grids and might not be apparent these functions
map to the CSS Grid Spec. As such, let's just
leave the documentation in the table of props
instead for the time being.


Jayphen:
Wait, what?

Why not just specify that these docs are for the CSS Grid spec, with a link to spec? Trying to 8 plement grids in a team using styled system is made difficult by the fact that the docs aren’t available

PR #152, opened 2018-03-09
Add examples to .npmignore

andrewreedy:
Examples folder is 172MB so it takes up a lot of space as an npm package which becomes an issue in certain environments.


johnotander:
Nice!

PR #115, opened 2018-02-19
WIP: Reorganize docs

jxnblk:

Issue #58, opened 2017-10-20
Using extras like borderRadius with components


babycourageous:
Hi there

Really enjoying using the style functions to unify the look of my site with these! But I'm having a bit of difficulty understanding how the extra props are used with existing components. In addition to importing any of them (i.e. `borderRadius`) I have included it as a prop in my component (a styled `Box` using `styled-components`. But there is no `border-radius` on my output. 

```
import styled from "styled-components";
import { Box } from "grid-styled";
import { space, borderRadius } from "styled-system";

const PerformanceTime = styled(Box)`
  display: inline-block;
  color: #fff;
  font-size: 12px;
  font-weight: 600;
  ${space};
`;

<PerformanceTime p={1} mb={1} borderRadius={1}>10:00pm</PerformanceTime>
```

I'm obviously missing something simple!
Thanks

gragland:
You need to include ${borderRadius} within your PerformanceTime style, like you did with ${space}.

babycourageous:
Thanks @gragland 

I originally had that as well. Tried putting it back just now but still nada. 
Should it just work or is there anything else I might need to do? All the base ones work great as is!

gragland:
@babycourageous Ah okay, not sure then as I haven't used `borderRadius` yet.

babycourageous:
@gragland thanks for your eye. Strangely I can't get any of those "extra" configurations to work on my components. Do I need a `styled-components` `Provider` with a `theme` prop by any chance wrapping up the App ?

gragland:
@babycourageous Just tested `borderRadius` and it looks like it expects a px/em/rem value. It doesn't read from theme spacing.

babycourageous:
Oh thanks for checking into that @gragland. That's too bad - was hoping to take advantage of some consistency there (like with `width` `margin` and `padding` props). 

Are you also a maintainer - both the `styled-system` and `rebass` docs and examples seem to have references to a `radius` or `radii` property in the documentation for theme and "extra props"

gragland:
Ah yes it looks like it gets values from `theme.radii` instead of `theme.spacing`. Maybe creating your own borderRadius style prop would do the trick (haven't tested this):
```js
import { responsiveStyle } from 'styled-system'
const borderRadius = responsiveStyle({
  prop: 'borderRadius',
  cssProperty: 'borderRadius',
  key: 'spacing'
})
```

babycourageous:
Thanks for all your help @gragland -- 

[...]

PR #156, opened 2018-03-14
document that borderRadius is responsive

ckeeney:
Document that `borderRadius` is responsive.

https://github.com/jxnblk/styled-system/issues/58#issuecomment-362851423.

This will likely cause a merge conflict with #115.

I think that #58 should be reopened until either #115 is merged with this change or this PR is merged.  I spent a while trying to get `borderRadius` to work and was only successful when I found the above comment.


jxnblk:
Sorry for the delay, but I don’t think borderRadius is Responsive at the moment...

PR #159, opened 2018-03-20
modify mx space to allow negative strings on the scale

mattywong:
This pull request allows using the negative values of a string based scale.

E.g

`
const space = ['1rem', '2rem', '3rem'];

//before
<Component mx={-1} /> // margin-left: 2rem, margin-right: 2rem;
// after
<Component mx={-1} /> // margin-left: -2rem, margin-right: -2rem;
`

### DIFF
--- a/src/space.js
+++ b/src/space.js
@@ -1,83 +1,79 @@
-import PropTypes from 'prop-types'
-import {
-  get,
-  arr,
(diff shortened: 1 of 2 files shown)


jasonblalock:
Any timeline on when this might get merged in?

PR #163, opened 2018-03-25
[System Classnames]: Allow control of vertical-align in tachyons

KingScooty:
I'm not sure if this package is only supporting the props that `styled-system` supports. If that's the case, feel free to close this.

However, i've been using this along with your `macro-components` package, and i've tried modifying more props, that this package doesn't offer.

This PR allows for changing vertical-align as a prop:

```
/**
v-base
v-top
v-mid
v-btm
*/

<div v={"-top"} />
```


jxnblk:
Thanks for this, looks great!

PR #165, opened 2018-04-05
Move getWidth to utils

johno:
Moving getWidth to utils will allow it to be
used by end users via the util export. This is
useful when you want to use getWidth for a different
style function. For example, defining the same
height as the width prop for an icon or avatar.

### DIFF
--- a/src/styles.js
+++ b/src/styles.js
@@ -3,13 +3,13 @@ import {
   style,
   pseudoStyle,
   responsiveStyle,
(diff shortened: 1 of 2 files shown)


jxnblk:
BTW, the `size` function is intended for width and height for square shaped elements like icons and avatars – it *might* make sense to use `getWidth` there as well, though percentage-based heights are generally a little less useful

Issue #150, opened 2018-03-08
Feature: Add ability to extend clean-tag blacklist by custom props


selrond:
Would it be possible to extend `clean-tag`'s blacklist with my custom prop / alias?

For example - I try to add `backgroundRepeat` prop function:

```js
const backgroundRepeat = responsiveStyle({
	prop: 'backgroundRepeat',
	alias: 'bgRepeat'
})
```

but since `bgRepeat` is not part of `clean-tag`s filtering blacklist, I'm getting 

```
Warning: React does not recognize the `bgRepeat` prop on a DOM element. ...
```

I know I could manually omit problematic props, but there has to be a better way - is there?

jxnblk:
You can pass your own array as the `blacklist` prop, but you'll need to include any other props you'd like to include along with it (i.e. it doesn't concat the default array)

selrond:
@jxnblk so wouldn't it make sense to improve this behavior?

jxnblk:
Perhaps. Not tested, but I imagine something like this would work:

```js
import styled from 'styled-components'
import tag from 'clean-tag'
import { backgroundImage, responsiveStyle } from 'styled-system'

const backgroundRepeat = responsiveStyle({
  prop: 'backgroundRepeat',
  alias: 'bgRepeat'
})

tag.defaultProps.blacklist = [
  ...tag.defaultProps.blacklist,
  'backgroundRepeat',
  'bgRepeat'
]

const Foo = styled.div`
  ${backgroundRepeat}
  ${backgroundImage}
`
```

Not opposed to something like a `customBlacklist` prop, but would want to figure out a good name for it... maybe it should accept a function?

samtietjen:
I also ran into this with system components. My work around for right now is:
```javascript
Button.defaultProps.blacklist = [   
  ...Object.keys(Button.propTypes),   
  'buttonSize'   
]
```
Which works, but can pollute props a bit. Is there a reason why blacklist isn't forced to concat with the propType keys?

ckeeney:
Here is a use case for this feature:

```js
import {Text} from 'rebass'

const MyText = Text.extend.attrs({blacklist: ['uppercase']})`
  text-transform: ${props => props.uppercase ? 'uppercase' : 'initial'};
`;
```

With the above code, the blacklist defined by Rebass is overwritten.

Kikobeats:
This definitely should be added into the doc!

jxnblk:
I think this should be resolved with #119

PR #170, opened 2018-04-24
Add a tip about warning

Kikobeats:
Reference: https://github.com/jxnblk/styled-system/issues/150


jxnblk:
Thanks! Do you think this should go in the readme for clean-tag instead? Worried that it might be easily missed in the main readme

Issue #162, opened 2018-03-23
[Question]: why is lineHeight not responsive?


harrisrobin:
Hello,

I really love this library, I have been using it on all my projects however I always find myself making a responsive lineHeight.. very useful and goes very well with fontSize={[...]}. Usually you need different lineHeights for different breakpoints so am I miss understanding something?

Thanks for everything, let me know if i can help with a PR!

harrisrobin:
@brandtnew i simply use `responsiveStyle`

```js
const lineHeight = responsiveStyle({
  prop: "lineHeight",
  cssProperty: "lineHeight"
})
```

jxnblk:
Related jxnblk/rebass#377

jxnblk:
It's not something I commonly use, but I'd accept a PR to make the lineHeight function responsive

jessehart:
i can see this being an issue if you're not using unitless line heights. but in that case, why?! :)

harrisrobin:
@jessehart I often follow designs from Sketch files that have very specific line heights on desktop and mobile, and i found that to get it pixel perfect I was better off simply copy pasting the "PX" value and adding breakpoints rather than trying to estimate it with a unitless line-height. Regardless, I think this lib should provide the option.

@jxnblk sounds good, ill see what i can do later today!

PR #173, opened 2018-04-28
make line-height responsive style

t49tran:
I ran into same problem at work as #162 and think it will a simple update.

In this PR:

- change `lineHeight` function from `style` to `responsiveStyle`
- update the test for `lineHeight` to test against responsive array
- update the docs to state that `lineHeight` is reponsive


test.js
```diff
@@ -957,17 +957,22 @@ test('lineHeight returns line-height', t => {
   t.deepEqual(a, { lineHeight: 1.23 })
 })
 
-test('lineHeight returns a scalar style', t => {
+test('lineHeight returns responsive line-height', t => {
```
jxnblk: Thanks for this! I'd prefer to keep the test that checks for the theme value as a separate one, like the other style functions. Would you mind updating this to include tests for both:
- `lineHeight returns a scalar style`
- `lineHeight returns responsive line-height`

t49tran: Yeah, that should be ok. I will include both test.

jxnblk:
Awesome, thanks!

Issue #169, opened 2018-04-12
Typescript support


chetmurphy:
Is it possible to add typescript support?

harrisrobin:
@chetmurphy right now it does not, however I did find and end up using this:

https://github.com/mujuni88/react-typescript-mobx-antd/blob/master/src/%40types/styled-system/index.d.ts

chetmurphy:
I found a different copy, but I am not convinced  that it is correct.  For example width is defined as

export function width(...args: any[]): any;

so when it is used

export const Container = styled.div`
  width: 100%;

then any value is valid. No type checking or autocomplete. It is also incomplete (does not contain all properties in styled-system).

To be fair there is also a typescript bug Allow type arguments in generic tagged templates #23430<https://github.com/Microsoft/TypeScript/pull/23430> that prevents overriding/merging types that is common in styled-system. A fix has been proposed but has not yet been integrated.

phobon:
So, I'm going to submit a PR with a more fleshed out version of what @harrisrobin linked... It works, but to use with a library like styled-components, you've got to use something like [styled-components-ts](https://www.npmjs.com/package/styled-components-ts). I've been using that combination like this:

````js
import styled from 'styled-components';
import styledTs from 'styled-components-ts';
import { width, space, WidthProps, SpaceProps } from 'styled-system';

type BoxProps = WidthProps & SpaceProps;

const Box = styledTs<BoxProps>(styled.div)`
   ${width}
   ${space}
`;

// You can apply typed default props here based on theme
Box.defaultProps = {
   m: 3,
   p: 2
};

// Then you can use typed props elsewhere (totally contrived), obviously
const BoxWithContent = (content: string): any => (
   <Box mr={2} pb={4}>{content}</Box>
);

export default BoxWithContent;
````

This gives all the static typing when composing, as well as when using the theme prop provided inline via something like: 

````js
width: ${props => props.theme.space[4]}px;
````

I'll submit the PR and see what people think, but it seems like it works well to me :)

phobon:
Also, to follow up maybe [this pull in ts2.9](https://github.com/Microsoft/TypeScript/pull/23430) (due this month) would solve some issues, so you could potentially do:

````js
import styled from 'styled-components';
import { width, space, WidthProps, SpaceProps } from 'styled-system';

type BoxProps = WidthProps & SpaceProps;

const Box = styled.div<BoxProps>`
   ${width}
   ${space}
`;
````

[...]

PR #181, opened 2018-05-12
Typescript definitions

phobon:
So, I've been using styled system with typescript and styled-components with this definition file. It was sourced originally from https://github.com/jxnblk/styled-system/issues/169 and extended to support all of the different functions.

I don't think it's perfect, but it's been working really well for me - just interested to see what everyone thinks. This is my first time contributing to something, I just love this project 👍

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,0 +1,398 @@
+declare namespace StyledSystem {
+  /**
+   * Core
+   */
+  export interface BaseTheme {
+    breakpoints?: number[];
+    space?: number[];
+    fontSizes?: number[];
+    colors?: {
+      [name: string]: string;
+    };
+  }
+
+  type ResponsiveValue<T> = T | (T | null)[];
+
+  /**
+   * Space
+   */
+  export type SpaceValue = number | string;
+  export type ResponsiveSpaceValue = ResponsiveValue<SpaceValue>;
+
+  export interface SpaceProps {
+    m?: ResponsiveSpaceValue;
+    mt?: ResponsiveSpaceValue;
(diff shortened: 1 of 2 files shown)


msokk:
Looks like styled-components will be moving definitions out of the repo to DefinitelyTyped in v4, maybe we should add these to DT too instead?

phobon:
I don't think that's a bad idea at all, probably makes sense to maintain things there as well

src/index.d.ts
```diff
@@ -0,0 +1,460 @@
...
+   * Utilities
+   */
+  export function theme(keys: string): any;
+  export function cleanElement(component: any): any;
+  export function removeProps(props: any): any;
+
+  /**
+   * Low level style functions
+   */
+  export interface LowLevelStyleFunctionArguments {
```
smashercosmo: 'alias' property is missing

jxnblk:
Thanks, but as I've discussed in many, many other issues, is there any reason to not host this in [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped)?

I don't use TS and won't be likely to keep this maintained over time

phobon:
That's all good - no reason at all why this can't live in DT. I'll make a PR there 👍

zephraph:
@phobon I've started a PR @ https://github.com/DefinitelyTyped/DefinitelyTyped/pull/26649. Do you want to be a maintainer on it? I've got you listed as an author currently.

phobon:
@zephraph Sorry about the late reply - thanks for setting that all up - looks really good! More than happy to be a maintainer 👍

PR #184, opened 2018-05-18
improve gitignore

selrond:
add ignore rules for macOS, Windows, Linux, Node


vadistic:
Btw. is there a reason why package-lock.json is being ignored?

selrond:
@vadistic no idea, @jxnblk?

jxnblk:
Thanks for the PR, but personally, I'd rather encourage folks to set global gitignores for editor and platform specific use cases. I think a project's gitignore should reflect only the build artifacts and other things that you might not want accidentally checked in. Hope that makes sense!

As for the package-lock ignore, I follow this approach: https://github.com/sindresorhus/ama/issues/479#issuecomment-310661514

PR #185, opened 2018-05-18
Make sure we're getting the value from an object and not string when getting styles from style object 

kimpers:
I have been having this issue where setting `width={1}` on a component gives the the width `0%`. 

I have managed to track down the issue and it seems that when getting the styles from the styles object it tries to get the key `1` from the string `"100%"` which happens to be `"0"`. Since `"0"` is a truthy value that is returned as the style setting. (See image below)

image

My proposed fix for this is to when reducing the style from the styles object check if a is an object and only then get they value of a[b] . If a is not an object and there is still path parameters left then there is no possible match for the full path so we should return `null` to use the fallback value instead.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -21,7 +21,7 @@ export const arr = n => Array.isArray(n) ? n : [ n ]
 
 export const getWidth = n => !num(n) || n > 1 ? px(n) : (n * 100) + '%'
 export const get = (obj, path, fallback) => path.split('.')
-  .reduce((a, b) => (a && a[b]) ? a[b] : null, obj) || fallback
+  .reduce((a, b) => (typeof a === 'object' && a && a[b]) ? a[b] : null, obj) || fallback
 
 export const mq = n => `@media screen and (min-width: ${px(n)})`
 


kimpers:
Might be tricky to get this reproducible with a small isolated example but I'll give it a go

jxnblk:
Are you attempting to use `width` with theme values?

kimpers:
Hi,

Sorry for not being very responsive with this. I've been busy with other things. Yes `with={1}` is used in conjunction with a general theme. The string `1000px` is coming from the site theme.

I don't think applying my fix is destructive in any way, I don't see any instance where you'd intentionally want to do `a[b]` where a is not an Object or Array.

jxnblk:
Can you provide a CodeSandbox or Gist of what your theme looks like?

kimpers:
I understand. I will try to dig out the code that was causing this when I have some time at work. We worked around it by setting `width="100%"` so I'll have to do some digging

kimpers:
I will close this PR as I haven't had time to look into establishing a repro case for the issue

PR #188, opened 2018-05-21
Fix innerRef for system-components

janv:
Solves jxnblk/rebass#329

StyledComponents doesn't properly recognize the generated wrapper and assigns it `ref`, when we actually want to pass through `innerRef`


jxnblk:
Thanks for this, looks great!

PR #191, opened 2018-05-27
Added `createSystem` export that allows to use `styled` from other libraries

vadistic:
Hello, I love `styled-system`, I'm a fan 👍 

But I also love `emotion` and was devastated to learn that in fact - I'm using `styled-components`  when using `system-components` 😭 

More serious: If your project already use other css-in-js library, you unnecessarily add `styled-components` to your bundle size (it gets even worse if you don't tree shake). Also, typescript definitions are work-in-progress but I'm sure I would quickly notice that my component type is incompatible with the `StyledComponent` type from i.e `emotion`.

So what I did:
- [x] It was just two lines of code to add createSystem export, that allows to create system interface that use any css-in-js library `styled` function.
```js
import { createSystem } from 'system-components'
import styled from 'react-emotion'

// creates a system function, that use `styled` from your library
const system = createSystem(styled)
```
- [x] Added test (it works, there's nothing to break- but I also tested it on some CRA with styled-components and emotion)
- [x] Added small section in readme (maybe could use a bit of polishing)

Hope you'll accept!

### DIFF
--- a/system-components/src/index.js
+++ b/system-components/src/index.js
@@ -1,8 +1,10 @@
 import styled from 'styled-components'
 import System from './System'
 
-const create = new System({
-  createComponent: type => (...args) => styled(type)([], ...args)
+export const createSystem = (styledFn) => new System({
+  createComponent: type => (...args) => styledFn(type)([], ...args)
 })
+
+const create = createSystem(styled)
 
 export default create
--- a/system-components/README.md
+++ b/system-components/README.md
@@ -113,6 +113,19 @@ const Bold = Text.extend`
 `
 ```
 
+## Usage with other css-in-js libraries
+
+By default, `system-components` use `styled-components` to when using default `system` export.
+
+You can create your own `system` interface that will only use your css-in-js library.
+
+```js
+import { createSystem } from 'system-components'
+import styled from 'react-emotion'
+
+// creates a system function, that use `styled` from your library
+const system = createSystem(styled)
+```
 
 ---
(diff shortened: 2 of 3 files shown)


vadistic:
I'm think now that the whole test suite should be run with `emotion`, (maybe something else?) to test compatibility, but I would need to add them as dev-dependencies. 

What do you think?

system-components/src/index.js
```diff
@@ -1,8 +1,10 @@
 import styled from 'styled-components'
 import System from './System'
 
-const create = new System({
-  createComponent: type => (...args) => styled(type)([], ...args)
+export const createSystem = (styledFn) => new System({
+  createComponent: type => (...args) => styledFn(type)([], ...args)
 })
```
jxnblk: Thanks for this! I was first considering exporting the `System` class itself, since it might eventually take more arguments that the `createComponent` argument. I think also having the empty array as the first argument is only necessary for styled-components, so I'd want to make sure that part is really customizable. The `createComponent` function could also be leveraged to do more than only swap out css-in-js libraries, if that makes sense...

jxnblk:
Heads up with #206 there's now a `system-components/emotion` module and the `System` class is exported for supporting other libraries (see https://github.com/jxnblk/styled-system/tree/master/system-components#using-with-other-css-in-js-libraries). 

Thanks for the initial PR, but I'm going to close this one out in favor of the new API

PR #196, opened 2018-06-06
Responsive size styles not being correctly merged in

allforabit:
The size style is not merging in media queries correctly because  it's not deep merging and so the media query property is just clobbered by the second (height) style. I noticed while using it with a responsive svg element.


jxnblk:
Thanks! Looks great!

PR #198, opened 2018-06-06
Added opacity style prop

jsonmaur:
Couldn't find anything in styled system for an `opacity` prop, so this adds it (as well as a few tests and updates to the documentation.) Also includes the shorthand `o`. This is a CSS value that is used a lot, so this could be really useful to have in the core.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -6,6 +6,7 @@
   - [width](#width-responsive)
   - [fontSize](#fontsize-responsive)
   - [color](#color-responsive) (and background-color)
+  - [opacity](#opacity)
 - [Typography](#typography)
 - [Layout](#layout)
 - [Flexbox](#flexbox)
@@ -157,6 +158,24 @@ Array values are converted into [responsive values][responsive-styles].
 
 // raw CSS color value
 <Box color='#f00' />
+```
+
+### opacity (responsive)
+
+
+```js
+import { opacity } from 'styled-system'
+```
+
+The opacity utility parses a component's `opacity` prop and converts it into a CSS declaration.
+
+```jsx
+// examples
+
+<Box opacity={0.5} />
+
+// with responsive values
+<Box opacity={[0.3, 0.7]} />
 ```
 
 ---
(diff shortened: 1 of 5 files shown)


docs/api.md
```diff
@@ -159,6 +160,24 @@ Array values are converted into [responsive values][responsive-styles].
 <Box color='#f00' />
 ```
 
+### opacity (responsive)
```
jxnblk: Thanks for the PR! I'd rather see `opacity` in the *Misc* section as I'd definitely not consider it a core part of a design system

src/styles.js
```diff
@@ -43,6 +43,11 @@ color.propTypes = {
   ...bgColor.propTypes,
 }
 
+export const opacity = responsiveStyle({
+  prop: 'opacity',
+  alias: 'o',
```
jxnblk: I think the shorthand should be removed; the existing shorthands are mostly meant for backwards compatibility, and I can't imagine `opacity` being used nearly enough to warrant a new shorthand

src/styles.js
```diff
@@ -43,6 +43,11 @@ color.propTypes = {
   ...bgColor.propTypes,
 }
 
+export const opacity = responsiveStyle({
```
jxnblk: What is the use-case for responsive opacity? I've never used opacity responsively, so forgive me if this is a silly question

jsonmaur: Don't have a good use case but seems like it could be used as much as responsive color or border would be. Not really sure what the criteria should be for a responsive style vs a regular style.

jxnblk:
Sorry for the delay on this, but thanks! Merging in and releasing shortly

src/styles.js
hnordt: @jxnblk @jsonmaur I think we should add `key: 'opacities'`, so instead of theming as `opacity: [0, 0.5, 1]` we would theme it as `opacities: [0, 0.5, 1]`. The plural form is the pattern Styled System follows for theming.

Issue #199, opened 2018-06-11
[Error] getting `TypeError: Cannot convert undefined or null to object`


selrond:
![image](https://user-images.githubusercontent.com/6603389/41218840-c4d80054-6d5c-11e8-999f-a6b64764da26.png)

After updating, I'm getting a TypeError. 

![image](https://user-images.githubusercontent.com/6603389/41218999-46b479cc-6d5d-11e8-997b-62893c2936d4.png)

Stack trace is quite unclear, but it's most probable to have something with #196

selrond:
Narrowed it down a bit more - commenting out 'size' prop function gets rid of the error:

```js
import system from 'system-components'

const Box = system(
  'space',
  'width',
  'fontSize',
  'color',
  // 'size',
)
```

allforabit:
I've added another pull request to fix this issue. To test the new version could you paste the following into `node_modules/styled-system/dist/styles.js` in place of the existing `size` function:
```
var size = exports.size = function size(props) {
  return (0, _util.merge)(sizeHeight(props) || {}, sizeWidth(props) || {});
};
```
I've added a default empty object to avoid the `null` error that you reported.

PR #200, opened 2018-06-11
Add default object to fix merge issue

allforabit:
As reported here https://github.com/jxnblk/styled-system/issues/199, the deep merge fix seems to have caused problems due to a null object being sent to `merge`. I've added default empty objects to avoid this.

### DIFF
--- a/src/styles.js
+++ b/src/styles.js
@@ -121,7 +121,7 @@ export const sizeHeight = responsiveStyle({
 })
 
 export const size = props => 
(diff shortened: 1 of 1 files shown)


jxnblk:
Thanks! Might be good to add a test for when size is null. Also, can this be addressed in the `merge` utility itself?

jxnblk:
Sorry, closing this in favor of #202

allforabit:
Great, that's probably the better solution alright to fix in `merge`.

PR #213, opened 2018-06-18
Add vertical-align

zephraph:
:wave:

Love this library. We've started using it at [Artsy](https://github.com/artsy) more heavily recently. This is hopefully the first of many contributions. 

I'm not exactly sure what category `vertical-align` should fall into. Also wasn't sure if it should be a responsive value. Didn't seem to me like it should be, but I'm more than happy to update it if it does.


zephraph:
I also might need to update the api docs... If layout is the right place for this property then I'll add it there.

zephraph:
@jxnblk Thoughts about the docs organization? I can move them if I need to. If not, I'll add the api section.

jxnblk:
Hey, thanks for this and sorry for the delay! I think *Layout* is fine for the docs. Merging and releasing shortly!

PR #214, opened 2018-06-19
Allow Alias for spacing

KeKs0r:
I am currently working on a design system that defines spacing like the following:

```js
export const space = [0, 2, 4, 16, 24, 32, 48, 64];
```

I would really like to access them via aliases e.g. 
```js
            <Box pt="xs" bg="primary">
              xs
            </Box>
```

With this PR, you can define aliases on your space scale. For example with this helper function
```js

const addAliases = (arr, aliases) =>
  aliases.forEach((key, i) =>
    Object.defineProperty(arr, key, {
      enumerable: false,
      get() {
        return this[i];
      }
    })
  );

export const space = [0, 2, 4, 16, 24, 32, 48, 64];
addAliases(space, ["none", "xxs", "xs", "s", "m", "l", "xl", "xxl"]);
```

### DIFF
--- a/src/space.js
+++ b/src/space.js
@@ -41,7 +41,7 @@ export const space = props => {
 
 const mx = scale => n => {
   if (!num(n)) {
-    return n
+    return scale[n] || n
   }
 
   const value = scale[Math.abs(n)] || Math.abs(n)


zephraph:
You can achieve this without a code change. Just change the array to an object. 

Here's how we define our space values... https://github.com/artsy/palette/blob/master/src/Theme.tsx#L104

And here's how it's used: 
https://github.com/artsy/reaction/blob/master/src/Styleguide/Pages/Artist/CurrentEvent.tsx#L33

KeKs0r:
@zephraph 

Thanks for the input. Unfortunately the "How its used" is not available to me.

I tried to get it working with looking at the theme.tsx, but it is still returning the alias, and not the aliased value.

This line: 
https://github.com/jxnblk/styled-system/blob/master/src/space.js#L43
is called with `n` being e.g. `xl`, but then it is checked that `xl` is not a number, and returns the string `xl` instead of the value with the key `xl`

zephraph:
Sorry about that, should've provided a perma-link. I'll update the links.

jxnblk:
Sorry for the delay on this, but this looks great! Thanks!

PR #219, opened 2018-06-22
Emotion should forward innerRef

tkh44:
Per this twitter conversation, https://twitter.com/colebemis/status/1009909682889035776, system-components was not passing innerRef through. This *should* fix it.

@mitchellhamilton or @Andarist might know of a better way


system-components/src/emotion.js
```diff
@@ -2,7 +2,8 @@ import styled from 'react-emotion'
 import System from './System'
 
 const create = new System({
-  createComponent: type => (...args) => styled(type)(...args)
+  createComponent: type => (...args) =>
+    styled(type, { shouldForwardProp: name => name === "innerRef" })(...args)
```
mitchellhamilton: It should be this.
```jsx
name => name !== 'ref' && name !== 'theme'
```

colebemis: Nice! I tried that out in a code sandbox and it worked 🎉 https://codesandbox.io/s/4qvpyo8r09

@jxnblk Looks like we didn't need to change `clean-tag` after all :)

tkh44:
This has been updated per @mitchellhamilton's fix [https://github.com/jxnblk/styled-system/pull/219/files#r197373080]

Should be good to go 👍

jxnblk:
Awesome, thanks!

PR #223, opened 2018-06-27
fix link

adekbadek:
link to "custom utilities" was broken because it lacked `.md` extension in file name

### DIFF
--- a/docs/custom-props.md
+++ b/docs/custom-props.md
@@ -1,7 +1,7 @@
 
 # Custom Style Props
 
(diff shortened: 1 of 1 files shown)


jxnblk:
This particular doc is used in the docs site, which doesn't currently strip out the `.md` when it's in the middle of the `href` prop. This will require an upstream change in x0 to parse out URLs with hashes like this

jxnblk:
Closing this out since the link works on the site, and I'll likely be updating the docs (hopefully handling this issue) soon

PR #224, opened 2018-07-02
Fix responsive-styles link

marhalpert:
https://github.com/jxnblk/styled-system/blob/master/docs/responsive-styles link is missing .md


johno:
Thanks!

PR #232, opened 2018-07-07
v3: performance rewrite

jxnblk:
This is an RFC for a potential major update to styled-system.

These changes aim to make styled-system:

- Smaller
- Faster (see [benchmarks][])
- Simpler API

Install the prerelease:

`npm i styled-system@alpha`

~~Note: the current version of styled-system is currently in the `ref/` directory for reference and benchmarking.~~

## Added

- `variant` new alternative to the current `complexStyle` utility
- `compose` composes multiple style functions into a single function
- `css` create style objects from props for all styled-system style functions

## Changed

- `style` utility is now responsive by default
- All built-in style functions now support responsive array values
- The style function `getter` argument has been renamed to `transformValue`
- `backgroundImage` no longer wraps values with `url()`
- ~~`themeGet` is now `get`~~

## Removed

- Prop aliases and the `alias` argument
- `propTypes[prop].meta` (temporarily removed)
- `responsiveStyle` use `style` instead
- `complexStyle` use `variant` instead
- `complexStyle` (`variant`) boolean props are no longer supported
- `pseudoStyle`
- `hover`
- `focus`
- `active`
- `disabled`
- `util.break`
- `util.media`
- `numberToPx` option (pass `util.px` to the `transformValue` argument instead)
- Main `propTypes` export (use style function propTypes instead)
- `borderWidth`

## Under consideration

- Is the `pseudoStyle` utility something worth keeping?
- Is the complex style boolean props feature worth keeping?

### `compose`

The `compose` function combines multiple style functions together.

```js
// example from src/styles.js
export const textColor = style({
  prop: 'color',
  key: 'colors',
})

export const bgColor = style({
  prop: 'bg',
  cssProperty: 'backgroundColor',
  key: 'colors'
})

export const color = compose(textColor, bgColor)
// <Box color='tomato' bg='black' />
```

### Removal of `responsiveStyle`

Many people were confused as to why only some style functions accepted responsive arrays as values. The new `style` function handles arrays like the current releases `responsiveStyle` and is faster than both the current `style` and `responsiveStyle` utilities.

### Removal of `complexStyle`

The name of this utility wasn't great, and the usage of boolean props for enumerable values is generally not a good idea. The new `variant` utility works in the same way as the current `complexStyle` but without support for boolean props. **If there is a strong desire to keep the boolean props API,** please let me know in the comments.

### Removal of `pseudoStyle`

The `pseudoStyle` API in the current release doesn't seem all that useful. While it attempted to map some CSS properties to theme values, it didn't handle all use-cases and generally seemed more confusing than helpful. **If there is a strong desire for this utility to not be removed,** please let me know in the comments, and I can investigate a better way to incorporate it into the library.

An alternative to the current `pseudoStyle` could be to use the new `variant` style function to get pseudoclass styles from a theme.

```js
// example of using `variant` for pseudoclasses
import { variant } from 'styled-system'

const theme = {
  hoverStyles: {
    primary: {
      '&:hover': {
        backgroundColor: 'blue'
      }
    },
    secondary: {
      '&:hover': {
        backgroundColor: 'gray'
      }
    }
  }
}

const hoverStyle = variant({
  prop: 'hover',
  key: 'hoverStyles'
})

const Button = styled.button([], hoverStyle)
// <Button hover='primary' />
```

### Removal of `numberToPx`

While it is a convenient argument to have, the same thing can be achieved by using the `util.px` utility in the `transformValue` argument.

```js
// old API
const fontSize = style({
  prop: 'fontSize',
  key: 'fontSizes',
  numberToPx: true
})

// new API
const fontSize = style({
  prop: 'fontSize',
  key: 'fontSizes',
  transformValue: util.px
})
```

### Removal of `util.media` and `util.break`

These were used internally, but don't offer a ton of utility outside of the library.

[benchmarks]: https://github.com/jxnblk/styled-system/tree/next-v3/bench

Bundle size:

```
 PASS  dist/index.js: 1.33KB (gzip)
 PASS  dist/space.js: 1.46KB (gzip)
 PASS  dist/styles.js: 1.89KB (gzip)
 PASS  dist/util.js: 1.49KB (gzip)
 PASS  dist/variant.js: 387B (gzip)
```


colepeters:
🎉 I'm so excited to use this! Thanks for your continued efforts, sir — component libraries would be a nightmare without Styled-system.

### Changed
> `themeGet` is now `get`

I'm curious about the rationale behind this. The `themeGet` name describes what it does perfectly, while `get` is generic. I understand that this change is likely due to the util's functionality having grown; however I'm also not clear on the new use case mentioned (that is, `const value = get(obj, 'path.to.nested.value', 'fallbackValue')`. Would love to hear the thinking behind this.

### Under consideration

> Should the `getter` argument be renamed to `transformValue`?

`transformValue` is approximately infinity% more clear to me than `getter`. Yes please!

> Is the `pseudoStyle` utility something worth keeping?

I've yet to come across a situation where I'd want to set a pseudo style on an instance of a component as opposed to within the component definition. I'm sure there may be valid use cases for this, but it seems like over-configuration. I feel the reduction in API surface area is a greater benefit. For the same reason, I'm 100% behind removing the `hover`, `focus`, etc utility props.

> Is the complex style boolean props feature worth keeping?

I generally feel like having two ways to refer to the same thing is just another opportunity for confusion (also for this reason, removing prop aliases feels like a good move). `variant` feels like exactly the right name to use, especially as `<Button variant='primary' />` makes things pretty much impossible to not understand. The marginal added verbosity (over `<Button primary />`) is worth it for the sake of consistency and clarity.

jxnblk:
Thanks for the thoughts/questions! As far as the `get` utility goes, I've essentially merged the two utility functions into one, where it checks the type of the first argument (https://github.com/jxnblk/styled-system/blob/next-v3/src/util.js#L25-L27). Currently, there is both a [`get`](https://github.com/jxnblk/styled-system/blob/master/src/util.js#L23) and a [`themeGet`](https://github.com/jxnblk/styled-system/blob/master/src/util.js#L189).This might be less clear, but seems like a nice API. I could definitely change my mind on this one.

As for `pseudoStyle`, I think it's best to deprecate unless someone has a good reason to keep it around.

And for the `complexStyle` to `variant` change, I've also considered adding a separate `booleanStyle` function for cases where a boolean prop might make sense.

colepeters:
Thanks for the follow-up! I have a couple further thoughts on the `get`, but first: do you have an example of how you see it being used with an object reference (as opposed to the theme)?

jxnblk:
I've decided to simplify the `get` function (since that seemed confusing). It now works exactly like the current `themeGet` utility. I'm still willing to consider a better name for it, and maybe it'll remain `themeGet`. It can be used like this:

```js
const Button = styled.button`
  background-color: ${get('colors.blue')};
`
```

sapegin:
Maybe just `theme` but then it's less clear that it's a function.

jxnblk:
I've reverted the get utility name to `themeGet`

PR #245, opened 2018-07-16
Convert alias value to PX

KeKs0r:
I added the capability, to have aliases for the spacing, but it also needs to be converted to pixels, if its a number. So it can be defined the same as without aliases:

e.g. 
```
export const space = {
  none: 0,
  0: 0,
  xxs: 2,
  1: 2,
  xs: 4,
  2: 4,
  s: 8,
  3: 8,
  m: 16,
  4: 16,
  l: 24,
  5: 24,
  xl: 32,
  6: 32,
};
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ package-lock.json
 .nyc_output
 coverage*
 site
+.vscode
--- a/src/space.js
+++ b/src/space.js
@@ -39,7 +39,7 @@ const getProperties = key => {
 
 const getValue = scale => n => {
   if (!num(n)) {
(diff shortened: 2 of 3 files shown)


jxnblk:
Thanks for the PR and sorry for the delay! Would you mind adding some unit tests for this functionality?

jxnblk:
Yeah, the tests are failing because of an upstream change in jest – I can push a separate PR to get those passing again

KeKs0r:
In that case, feel free to merge this or wait for the upstream fix :)

PR #252, opened 2018-07-22
Add system-component blacklist documentation

flybayer:
Let me know if you want any changes!


jxnblk:
Thanks for the PR! But I don't think that I want to support the `blacklist` prop as public API. It *might* work, but I wouldn't rely on it. I think this is an issue that should be fixed upstream in the styled-components project, or if you're defining a lot of custom style props, you can try using the `clean-tag` package instead

PR #255, opened 2018-07-29
Add some missing style props

### LINKED ISSUE #240: Support justify-items
Thanks for all the thoughtful effort here!

I was working today and found support for the `justify-items` property is missing. Perhaps it is intentional or I am confused somehow, but I thought it worth flagging.

flybayer:
This adds `fontStyle`, `overflow`, `justifyItems`, `gridTemlateAreas`, and `gridArea`.

Closes #240 

If you want me to rebuild the docs for this PR, I can do so.

### DIFF
--- a/docs/table.md
+++ b/docs/table.md
@@ -75,6 +75,7 @@ Function Name | Prop       | CSS Property    | Theme Field
 `gridAutoColumns` | `gridAutoColumns` | `grid-auto-columns` | none
 `gridTemplateRows` | `gridTemplateRows` | `grid-template-rows` | none
 `gridTemplateColumns` | `gridTemplateColumns` | `grid-template-columns` | none
+`gridTemplateAreas` | `gridTemplateAreas` | `grid-template-areas` | none
 
 ## Background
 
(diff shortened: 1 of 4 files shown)


jxnblk:
Thanks for this! I think the tests are only failing because of an update in Jest that essentially causes all the tests to fail. (The solution is adding `"testURL": "http://localhost"` to the jest config – I can handle this later)

In order to get this in, the docs will need to be updated as well, if you wanna take a stab at that

flybayer:
@jxnblk I added docs and fixed the tests!

docs/table.md
```diff
@@ -68,6 +69,7 @@ Function Name | Prop       | CSS Property    | Theme Field
 `gridColumnGap` | `gridColumnGap` | `grid-column-gap` | `space`
 `gridColumn` | `gridColumn` | `grid-column` | none
 `gridRow` | `gridRow` | `grid-row` | none
+`gridArea` | `gridArea` | `grid-area` | none
```
Lavoaster: I think you're missing `gridTemplateAreas` here.

system-components/package.json
```diff
@@ -44,6 +44,7 @@
     "roots": [
       "<rootDir>"
     ],
+    "testURL": "http://localhost",
```
Lavoaster: Jest `23.5.0` came out a few days ago which resolved the need to have this in the config. But it's also harmless to have.

docs/table.md
flybayer: You are right, thanks for the catch!

Lavoaster:
@jxnblk Is there anything preventing this from being pulled in?

PR #258, opened 2018-07-31
Provide cjs and esm bundles

TrySound:
I'm trying to build my app with rollup and its commonjs plugin is not
able to convert weird babel output to esm because of `undefined` resets.

https://unpkg.com/styled-system@3.0.2/dist/index.js

In this diff I provide flat bundles for both cjs and esm. The biggest
benefit for me is hidden internals so api can be under better control.

Added also my size-snapshot plugin which also provides information about
treeshakability.


TrySound:
Tests are probably failed because of some dependency patch. Don't you like lock files?

jxnblk:
Thanks for this! I'll try to take a look at this over the weekend. One thing I am curious about (if you have an answer) is what are the pros/cons of having a setup like this versus using a tool like [microbundle](https://github.com/developit/microbundle)?

TrySound:
Well, rollup out of the box requires some config. But this config is quite small. microbundle provides too much, maybe not even necessary stuff. It's a big load in your node_modules.

Size snapshot also provides more info about your bundle.

jxnblk:
I don't really care how large a dev dependency is (within reason of course), but seems like not needing to mess with a config is a compelling reason to use it... Do you have any sense of how the output might differ?

TrySound:
I don't know. I prefer to solve problems with simple config. This allows me to not stick with somebody's constraints. At least size snapshot with treeshakability info is not produced by `microbundle`.

TrySound:
One more argument. microbundle uses buble which worked on in the past. I know how it's complex and error prune. It's syntax is quite limited. Babel is more universal and stable.

jxnblk:
Thanks for all the work on this and other projects! Closing this in favor of #284, please let me know if there's anything missing from that approach

Issue #257, opened 2018-07-30
Supporting objects instead of arrays for responsive props and scales


johanneslumpe:
The issue with arrays is that while they provide an implicit order to breakpoints, they are brittle to use in practice - especially in large applications with changing style-guides or new sizes that need to be accommodated.

Objects on the other hand don’t provide this implicit order but are safe to use even when new breakpoints or spacing values are inserted between existing values.

While non-implicit order in an object requires either good names or good documentation (the latter can easily be solved in TypeScript using interface comments and editor hits) it provides a much more resilient way to specify values which are safe from future disruption - short of changing an existing value.

Would you consider adding support for objects or switching from arrays to objects entirely? I’d be open to submit a PR for this if if has good chances of getting merged.

It's completely fine if this is not something you are looking to do, but I wanted to reach out before either forking or creating something similar which is more focused on these requirements.

Either way: styled system itself is a great library, keep up the good work!

hamlim:
+1 to this issue, internally we looked closely at styled-system, and we may adopt it in the future, however we worried about using arrays to represent responsive props would be brittle to updates to our design system tokens.

We decided to opt for our own implementation that uses objects with keys representing the breakpoint, which looks like this:

```jsx
<Block
  width={{
    [BREAKPOINTS.medium]: 50%
  }}
/>
```

IMO it would be nice to support both, so those that are currently using arrays will still work and might not make this a breaking change.

jxnblk:
Thanks! This definitely comes up in conversation from time to time. I think I'd first like to see a proposal of how it could be implemented, but I think it could potentially be a completely different set of functions for people who want the different api. I would definitely keep maintaining the array based props since I think that it's a nice constraint to work within

johanneslumpe:
@jxnblk sounds good! I’ll try to open a proof-of-concept PR in the next days!

jxnblk:
It sounds like this might be better handled by another library since it's quite different from the current API. Going to close the issue out for now, but feel free to open a new one if needed

joltmode:
This has been introduced with https://github.com/jxnblk/styled-system/pull/341

PR #259, opened 2018-08-01
initial proof of concept for responsive object values

johanneslumpe:
@jxnblk This is just a very quick proof of concept for https://github.com/jxnblk/styled-system/issues/257 to show that it's possible to do this inline within the existing code. We can see that we now have quite a few checks around whether or not the value is an object. It does pass all tests though. This should be a good starting point for a general discussion whether or not this should be inline or broken out into a separate piece of functionality.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -15,6 +15,11 @@ export const propTypes = {
 }
 
 export const defaultBreakpoints = [ 40, 52, 64, ].map(n => n + 'em')
+const defaultBreakpointsKeys = ['small', 'medium', 'large']
+export const defaultBreakpointsObject = defaultBreakpoints.reduce((acc, val, index) => {
+  acc[defaultBreakpointsKeys[index]] = val
+  return acc
+}, {})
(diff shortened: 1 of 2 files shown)


johanneslumpe:
cc @hamlim

johanneslumpe:
The test failure is related to an issue with JSdom I believe and not to the actual code change.

johanneslumpe:
@jxnblk did you have some time to think about this? :)

jxnblk:
Planning on taking a look at this and some of the other styled-system PRs/issues this weekend, haven't had a tone of time this week, sorry

johanneslumpe:
No worries, just checking in!

jxnblk:
Thanks for the PR! This seems like more conditional logic than I'd like to introduce into the core library. I do still think there might be room for a rewritten separate package to handle object values instead of arrays, if that's something you want to explore further...

johanneslumpe:
Oh absolutely! Ok definitely going to work on something separate then :)

Issue #241, opened 2018-07-13
Bug: system component's extending component won't apply blacklist option


youngboy:
for the case like in docs http://jxnblk.com/styled-system/system-components#extending-components

add blacklist option won't have any effect for rendered DOM, this will causing some React props warning in some cases..

ideally, I could just using emotion's `shouldForwardProp`, but seems I am not able to configure this since it is hard-coded https://github.com/jxnblk/styled-system/blob/master/system-components/src/emotion.js#L7

according to  https://github.com/jxnblk/styled-system/blob/master/system-components/src/System.js#L61, clean-tag is not apply when `is` prop is function.

for now I just added the following code to my project as a simple workaround...

```
import { omit } from 'clean-tag'

const cleanCmp = (Cmp) => {
  return ({blacklist, ...props}) => (
    <Cmp {...omit(props, blacklist)} />
  )
}

Link = system({
  is: cleanCmp(RouterLink),
  blacklist: [
    'is'
  ]
},
  // more ...
)

```

frontendmonster:
Same here after upgrading to 2.3.6. 
It says: `Warning: Received a 'function' for a string attribute 'is'. If this is expected, cast the value to a string.`

and it passes all props to the dom (even blacklist).

```html
<svg class="sc-bZQynM bmMrtV" color="colors.secondary.light" blacklist="innerRef,css">
```

youngboy:
The bug still exist in v3.0.2,

furthermore, in v3.0.2  user defined blacklist option won't merge with to builtin blacklist settings, see
https://github.com/jxnblk/styled-system/blob/master/system-components/src/System.js#L90

```
      Component.defaultProps = {
        ...baseProps,
        blacklist: [
          ...blacklist,
          ...(baseProps.blacklist || [])
        ],
        ...defaultProps // defaultProps.blacklist will override previous value
      }
```

jxnblk:
I didn't intend for the `blacklist` prop in system-components to be part of the public API. This is an issue with the styled-components library, that should probably be resolved there. There is a long-running issue there to track progress: https://github.com/styled-components/styled-components/issues/439

If you want to use the blacklisting feature, I'd recommend using [clean-tag](https://github.com/jxnblk/styled-system/tree/master/clean-tag), as is mentioned in the docs. Even if the above PR is merged, I wouldn't recommend relying on the `blacklist` prop to work in system-components

flybayer:
Yeah @jxnblk can you help us understand this a bit more?

[...]

PR #260, opened 2018-08-03
System component blacklist fix

youngboy:
This PR is tried to resolve https://github.com/jxnblk/styled-system/issues/241

Also note that I added one more test case for extend non-system component. Pass the non-system component to defaultProps still failed. I could fix this, but the fixes will break another test 
```javascript
  test('extends a non-system component and does not accept an is prop', ...)
```
I am not sure whether it is designed by purpose this way.. IMHO this test should be removed ...

right now to extend component, best pass custom component during rendering rather than default option, and everything works fine
```jsx
render(<Ext is={(props) => <div {...props} />} customProp='hi' bg='tomato' />)
```

### DIFF
--- a/system-components/package.json
+++ b/system-components/package.json
@@ -41,6 +41,7 @@
     "styled-components": "^3.3.2"
   },
   "jest": {
+    "testURL": "http://localhost/",
     "roots": [
       "<rootDir>"
     ],
--- a/system-components/src/System.js
+++ b/system-components/src/System.js
@@ -83,11 +83,11 @@ class System {
       const baseProps = util.get(defaultProps, 'is.defaultProps') || {}
       Component.defaultProps = {
         ...baseProps,
+        ...defaultProps,
         blacklist: [
           ...blacklist,
           ...(baseProps.blacklist || [])
         ],
-        ...defaultProps
       }
       Component.propTypes = propTypes
       Component.systemComponent = true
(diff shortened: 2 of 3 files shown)


Kikobeats:
Loving if @jxnblk can merge this 🙏

jxnblk:
Thanks for the PR!

> I am not sure whether it is designed by purpose this way

This would break quite a few dependents of this library, but yeah it was a trade off for how extending and the `is` prop work and I think this should be a different issue altogether. Also I’m unsure of whether the `blacklist` prop should even be part of the public API at all

youngboy:
my use case for `blacklist` is when I need to set conditional stylings base on props.

etc, `active` props in `<Link active />`

yes, we could archive the same result by using `variant` style. but I think sometime using dynamic stylings inline feels easier ...

flybayer:
I'm with @youngboy — I **really** like having the blacklist option for the cases he described.

ronnyhaase:
My 2 cents are that I also have a strong need for this feature, for exactly the reason @youngboy pointed out [before](https://github.com/jxnblk/styled-system/pull/260#issuecomment-410244083)

jxnblk:
I've moved and renamed system-components to a new repo here: https://github.com/rebassjs/components

The new version might address some of the issues here, but future development should continue in that repo, leaving this one focused on styled-system core

PR #267, opened 2018-08-18
Refactor system-components

### LINKED ISSUE #265: Question: How to extend react components in system-component@3.0.0?
Hi, I have a question.
with `system-components@2.2.2` and I had a Link system component that extends from `react-router` NavLink and it was working as expected.

```javascript
const Link = system({
  is: NavLink
  }
  //...
);
```

then, I upgrade `system-components` to v3.0.0 and now I'm getting a `Warning: Received a 'function' for a string attribute 'is'. If this is expected, cast the value to a string.` error and it stops working as Link component.

is this bug or we should just use HTML tag as `is` prop?

jxnblk:
This is an attempt to simplify the codebase and API of system-components and would be a major change.

New API:

```js
system(defaultProps, ...functionNamesOrStyles)
// returns StyledComponent
```

Example:

```js
const Box = system({
  is: 'section',
  blacklist: [
    'customProp'
  ],
  p: 3,
  bg: 'tomato'
},
  'space',
  'color',
  'font-weight:bold;',
  { textDecoration: 'underline' }
)

const Flex = system({
  extend: Box
},
  { display: 'flex' },
  'alignItems',
  'justifyContent'
)
```

## Changes

- Adds `extend` prop for extending other system-components
- The `is` prop should no longer be used to extend components
- The `is` prop has been renamed to `tag` for changing the underlying HTML element
- The `is` prop is deprecated, but will continue to work for changing the HTML element
- The first argument **must be** a `defaultProps` object
- styled-system functions are no longer inferred from the `defaultProps`
- Adds support for custom `blacklist` prop – before this only worked in some cases
- Changes build setup for emotion support
- Removes `System` class

See also:
- Fixes #265, jxnblk/grid-styled#137, jxnblk/grid-styled#127
- Related: #241, #260, jxnblk/grid-styled#135, jxnblk/grid-styled#95


flybayer:
Nice! Here's my thoughts/questions:

1. I think changing to `extend` and `tag` is great. Today I was running into an issue where I wanted to extend the styles from a base component but change the tag. *So this change should allow this, right?*
```js
const Textarea = system({ tag: 'textarea', extends: BaseInput })
```
2. *Why the change to not infer styled-system functions from the `defaultProps`?* I think the current behavior is very handy, but if this signifigantly improves something like performance, then I'd say it's worth the change.
3. *Does `functionNamesOrStyles` still include custom style functions?* Like this:
```js
const Box = system({ 'space', myCustomStyleFunc })
```
4. Hurray for `blacklist`!! 😍 

Thank you for your continued work on this!

flybayer:
I've felt a bit of what you are referring to @shawnbot where, as a beginner it's hard to figure out what's really going on. But I'm not sure I like your proposed solution.

Here's another idea
```js
const Button = system(
  {
    tag: 'button', 
    extends: Base, 
    defaultProps: {m: 0, p: 0}
  }, 
  'flex'
)
```

jxnblk:
@flybayer thanks! I'll try to explain some of the rationale

> Why the change to not infer styled-system functions from the defaultProps? I think the current behavior is very handy, but if this signifigantly improves something like performance, then I'd say it's worth the change.

The change is mostly to make the API less magical and more intentional. I know the current API is convenient, but it also adds more complexity to this module

> Does functionNamesOrStyles still include custom style functions?

Yes. Essentially anything that styled-components or emotion can handle will work for these arguments.

jxnblk:
@shawnbot thanks! I certainly don't disagree with the fact that the `props` argument is a little overloaded, but trying to keep this somewhat inline with the current API. As for the `is` and `tag` props, I *do* think they are very useful as props because they allow usage like this: `<Box is='footer' />`, and I know a lot of people rely on that functionality. That said, the `extend` prop might make more sense to be hard-coded into the API here since that is unlikely to change when using the component

jxnblk:
Heads up, I've moved and renamed system-components to `@rebass/components` here: https://github.com/rebassjs/components

The v4 beta includes the work here and I will be deprecating the system-components package

Issue #269, opened 2018-08-21
mistake in semver notation within package.json for clean-tag ?


jsms90:
In clean-tag's `package.json`:
> "styled-system": ">=2.0.0 || >=3.0.0"

But I'm guessing this is a mistake, as "greater than or equal to 2" _or_ "greater than or equal to 3" is basically the same as just saying `>=2.00`. Unless I've misunderstood something?

I mention this in particular because we seem to be experiencing problems with using `grid-styled` for layout, when we update our dependencies & get version `3.0.0` of `styled-system` - see this issue in our application: https://github.com/elifesciences/elife-xpub/issues/500

jxnblk:
As mentioned in the PR, clean-tag should work for either version. If there's an issue with grid-styled, please open an issue here https://github.com/rebassjs/grid – I'll be looking into resolving some of the dependency issues soon

PR #270, opened 2018-08-21
Depend on < v3 of styled-system for new v1.1.1

jsms90:


jsms90:
Oops sorry. That was meant to be a PR to my fork. Just playing around to see if I could sort #269, at least within our app for now.

jxnblk:
Assuming I should close this out? I think that clean-tag should support both v2 and v3. That said, grid-styled probably needs to be updated to resolve any issues there

Issue #142, opened 2018-03-02
Feature: styled-components/primitives


benjamingeorge:
Is it possible to use system-components to create Text and View components from the styled-components/primitives package?  I'm trying to make a styleguide that renders to both web and react-sketchapp.

jxnblk:
It should work but I've never tested it. I assume the breakpoints would not work, since that's not something that exists in react-native

okonet:
I’m wondering if we could have a separate entry point for that. I’d like to experiment with universal rendering as well.

johno:
I def think we should implement a way to support primitives as a separate entrypoint. I'm going to reopen this issue @benjamingeorge since it's something we'd like to put on the roadmap (whether as a separate entypoint or package in the monorepo).

johno:
Do you want to spike out a proof of concept @okonet?

okonet:
I could actually. No ETA but let me try soonish

johno:
Cheers! Yeah, no rush. Let us know when you have something. And feel free to open up an early WIP PR 💟 .

jxnblk:
FWIW system-components uses a `createComponent` argument that passes styled-components to the main class. This should be written in a way to support other libraries like `styled-components/primitives`, glamorous, emotion, etc.

see https://github.com/jxnblk/styled-system/blob/master/system-components/src/index.js#L5

okonet:
Yep, that makes sense. Also I was thinking about simple aliasing it. Passing the factory function is definitely cleaner but I’m wondering how this will look like in the user land. Ideally, users shouldn’t change any of the components code to do that.

allforabit:
I have started experimenting with using styled system with React Native and React Native Web. In general it seems to work quite well. One of the issues that has cropped up so far is that a lot of the styles add px which won't work with react native and requires unitless values. To make it work I've made the following tweak:
```
export const isReactNative = () => typeof navigator != 'undefined' && navigator.product == 'ReactNative'
// ...
export const px = n => isReactNative() ? n : (num(n) ? n + 'px' : n)
```

It just checks to see if it is running in React Native and if so doesn't add the 'px'. 

As React Native doesn't use css, you don't really need a css-in-js library and I've been using a higher order component that uses the new React context system to add the styles and theme:
```
const Context = createContext();

export const ThemeProvider = Context.Provider;

export const styled = (...styles) => Comp => {
  // Render function
  return ({ style = {}, ...props }) => {
    return (
      <Context.Consumer>
[...]

PR #279, opened 2018-09-06
React Native: Avoid adding px to unit values

kevinwolfcr:
# What
Removes `px` units when using with React Native.

# Why
React Native does not accept `px` as unit values.

screen shot 2018-09-05 at 19 09 14

# How
Based on [this comment](https://github.com/jxnblk/styled-system/issues/142#issuecomment-386592680) from #142, I've just added an environment check on the `px` function and if the user is using React Native, then return without adding the `px` suffix.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -14,10 +14,11 @@ export const propTypes = {
   ]),
 }
 
+export const isReactNative = () => typeof navigator != 'undefined' && navigator.product == 'ReactNative'
 export const defaultBreakpoints = [ 40, 52, 64, ].map(n => n + 'em')
 export const is = n => n !== undefined && n !== null
 export const num = n => typeof n === 'number' && !isNaN(n)
-export const px = n => num(n) ? n + 'px' : n
+export const px = n => (isReactNative() ? n : num(n) ? n + 'px' : n)
 
 export const get = (obj, ...paths) => paths.join('.').split('.')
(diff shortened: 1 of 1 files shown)


jxnblk:
Thanks! I don't think I'd want to add conditional logic in the runtime for something like this. Would you be willing to look into a babel-based solution? i.e. something where a build is compiled during publish to support this?

kevinwolfcr:
I've never implemented a babel solution that solves that issue, do you have any example for that so I can use it and adapt?

jxnblk:
I don't know of anything off the top of my head. It would be similar to how you've implemented it, but change the code to either include or not include the `'px'` string concatenation during the `prepare` step – I think a good first step would be to search for a babel plugin that can pre-evaluate the code in the `px` function based on the environment

kevinwolfcr:
Got it. Still, I have the doubts of why you don't want to include 1 line of code that can help **_a lot_** of developers that may be experiencing the same issue I have on the future?

jxnblk:
It's a trade-off. I *do* want this to be compatible with RN and useful to those developers, but I don't want added complexity in the runtime part of the codebase to support things like this. The `px` might be the only thing that needs changed, but there's also a lot of code in styled-system for media queries that is completely unnecessary for RN, and I'd rather have something set up to support changes like that when needed so that different platforms have more optimized/custom tailored options

PR #282, opened 2018-09-06
Move all modules to a single file

jxnblk:
Moves all source code into a single file, for my sanity.

@TrySound I'm curious to know if this is a viable option to the problem that #258 attempts to solve (with the addition of some babel config changes and updates to the `prepare` step) – also riffing off of what you did in rebassjs/components#2

I'll likely merge this in regardless, but to clarify the hesitation with adding rollup to *all* the packages I maintain, I see it like this:
- Pure Node.js modules didn't use to require a build setup
- Babel helps make node modules compatible with browsers
- Adding another dev dependency and config (on top of babel) for small packages is a lot of overhead
- I think that dependencies should allow consumers to handle their own build setup however they like
- I want to support tree-shaking and help consumers create the smallest possible bundles they can
- I want a standardized way to support ES modules and tree shaking across all my projects, if possible

Ideally, packages like this could be written in a standard that wouldn't require any build setup at all, but I understand that we aren't there yet. Any guidance on this would be much appreciated


sapegin:
Correct me if I'm wrong, wouldn't it prevent tree shaking because now everything is in one module?

TrySound:
@sapegin Importing from different files is not tree shaking. Bundler is smart enough to omit unused stuff from each module.

TrySound:
@jxnblk I'm okay with this if you don't want to introduce rollup. However size-snapshot plugin is necessary to test treeshakability you want to achieve. If you want I could release standalone version.

jxnblk:
Yeah, a dev tool for testing treeshakability is a nice to have

> If you want I could release standalone version.

^^ by this do you mean add a plain Node/CLI API to the plugin? or did I misunderstand?

TrySound:
Yes just a simple command which will wrap rollup+size snapshot plugin

TrySound:
@flybayer microbundle is rollup but with some opinionated stuff included.

jxnblk:
@flybayer Yeah, I've tried microbundle (like it conceptually) but doesn't seem to work super well across the packages I've tried it on

PR #287, opened 2018-09-13
Distribute esm with module field

TrySound:
We found some interop problems with `mjs` dual mode. It's probably
better to be consistent and solve the problem with `module` field
like I did in rebass.


jxnblk:
Thanks! Yeah, I think since this doesn't have any hard dependencies, the webpack issue wasn't happening here, but good to be more consistent

package.json
```diff
@@ -33,6 +34,7 @@
   "devDependencies": {
     "@babel/cli": "^7.0.0",
     "@babel/core": "^7.0.1",
+    "@babel/plugin-transform-runtime": "^7.0.0",
```
jxnblk: This will require end-users to have babel runtime installed as well, right? I thought that it wasn't good practice to use this in an npm package

TrySound: This is actually is the best practice since it allows to share the same helpers with other packages. And if user also have babel then his helpers will be also reused.

TrySound:
So, do you want me to rebase this PR?

jxnblk:
Hey sorry, ran into a bunch of bugs last night with 3.1, working on resolving them and I'll get the conflicts here resolved an merged in once I know what's going on. Tracking it in #288 – I think it was related to the single file move, but trying to isolate the changes. If you're using styled-system, I'd say stick with `3.0.4` for now

package.json
TrySound: Done

jxnblk:
Thanks, I *think* 3.1.2 fixes the issues I was seeing so going to merge this in and test it out across a few dependent libraries as well

TrySound:
@jxnblk You can use `squash and merge` button to prevent commits noise in your history. There is also configuration to prevent accidental using of any other buttons.

jxnblk:
Thanks, I think I have squash disabled by default on most of my repos

PR #292, opened 2018-09-19
Handle falsey values in arrays

johno:
v3 doesn't account for null usage as values for
responsive props like the following:

```js
<Box p={[null,1,2]} />
```

This ensures that when a null value is passed we
default to an empty object.

---

Something to consider: would it be cleaner to modify the `style` function to return `{}` rather than `null`? I went back on forth on where to handle this logic.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2744,7 +2744,7 @@
     },
     "@sinonjs/formatio": {
       "version": "2.0.0",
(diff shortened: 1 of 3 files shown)


jxnblk:
I *think* that returning `null` will ensure that styled-components/etc. won't create an empty CSS rule when there aren't any styles, but this would need to be confirmed

PR #294, opened 2018-09-21
🐛 Default media query to px if breakpoint value is unitless

### LINKED ISSUE #248: Units of breakpoints
Since I upgraded the version to 3 of `styled-system` I see inconsistencies among deprecated `responsiveStyle` and `style` utility. Image the following breakpoints defined by the theme:

```js
[ 480, 768, 992, 1200, 1550 ]
```

How to make it work with `pixels` and not `em`s? Because from now on I see media queries are rendered without any units like:

```css
@media screen and (min-width:480) {
  .myElement {
    text-align: left;
  }
}
```

BrendanFDMoore:
Fix #248 

The PR restores `createMediaQuery` behaviour to leverage `px` helper to supply default units as `px` when breakpoint value is unitless.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -48,7 +48,7 @@ export const compose = (...funcs) => {
   return fn
 }
 
-export const createMediaQuery = n => `@media screen and (min-width: ${n})`
+export const createMediaQuery = n => `@media screen and (min-width: ${px(n)})`
 
 export const style = ({
   prop,


BrendanFDMoore:
I just pulled this out of the source from v2 latest and dropped it in the v3 code. Let me know if you removed this for a specific reason and were intending to solve another way.

BrendanFDMoore:
```
var Benchmark = require('benchmark');
// var Fns = require('../src/index');

const num = n => typeof n === 'number' && !isNaN(n)
const px = n => num(n) ? n + 'px' : n
const createMediaQuery = n => `@media screen and (min-width: ${px(n)})`
const createMediaQueryPrev = n => `@media screen and (min-width: ${n})`


const suite = new Benchmark.Suite;

// add tests
suite.add('createMediaQuery', function() {
  createMediaQuery(680);
  // createMediaQuery('680');
  createMediaQuery('680px');
})
.add('createMediaQueryPrev', function() {
  createMediaQueryPrev(680);
  // createMediaQueryPrev('680');
  // createMediaQueryPrev('680px');
})
// add listeners
.on('cycle', function(event) {
  console.log(String(event.target));
})
.on('complete', function() {
  console.log('Fastest is ' + this.filter('fastest').map('name'));
})
// run async
.run({ 'async': true });
```
👇 
```
➜  styled-system git:(bm-fix-248-breakpoint-units) ✗ node ./scripts/benchmark.js
createMediaQuery x 15,377,376 ops/sec ±1.26% (82 runs sampled)
createMediaQueryPrev x 54,337,096 ops/sec ±0.63% (88 runs sampled)
Fastest is createMediaQueryPrev
➜  styled-system git:(bm-fix-248-breakpoint-units) ✗ node ./scripts/benchmark.js
createMediaQuery x 27,411,604 ops/sec ±0.94% (81 runs sampled)
createMediaQueryPrev x 55,912,352 ops/sec ±0.37% (84 runs sampled)
Fastest is createMediaQueryPrev
➜  styled-system git:(bm-fix-248-breakpoint-units) ✗ node ./scripts/benchmark.js
createMediaQuery x 60,470,792 ops/sec ±1.22% (83 runs sampled)
createMediaQueryPrev x 602,038,440 ops/sec ±0.39% (86 runs sampled)
Fastest is createMediaQueryPrev
```

It's certainly slower when it has to check the type and add the suffix, but also resolves the linked issue.

Alternatively, we could simply update the theming docs to be explicit that the breakpoint values should carry a unit now due to performance optimizations.

jxnblk:
Thanks! I was mostly curious to see benchmarks against the currently documented ones, but this is helpful.

I'm wondering if supporting this is even worth it, since `createMediaQuery` is called in the `style` function. I don't think having to use strings with unit values is too crazy (since it's in one place in your theme)

> Alternatively, we could simply update the theming docs to be explicit that the breakpoint values should carry a unit now due to performance optimizations.

Let me think about it, but I think this option might be the better way to go

aczekajski:
Since it's for performance reasons, but at the same time this is a regression, I'd like to suggest turning this  behaviour into opt-out thing. For example, my project reuses the same `breakpoints` from `theme` for other mechanisms and relies on them being numbers, not strings (for the sake of making calculations on them).

My suggestion is to have an additional `breakpointsUnit` field into `theme`. If it's not falsy, it will be simply concatenated inside `createMediaQuery`:
```
createMediaQuery = (n, unit) => `@media screen and (min-width: ${n}${unit || ''})`
```

BrendanFDMoore:
Oops @jxnblk , I missed this:
> see benchmarks against the currently documented ones
where can I see the current ones and how can I provide the same numbers?

@aczekajski That could be a reasonable compromise, which I can test for perf based on the above. The extra check alone will incur _some_ extra cost, it's just a question of how much.

BrendanFDMoore:
Ok, got `npm run bench` to work eventually by renaming the built cjs file to `index.js` in `/dist`.
*Benchmarks with this change:*
```
space x 457,043 ops/sec ±0.80% (88 runs sampled)
width x 1,699,514 ops/sec ±0.47% (91 runs sampled)
fontSize x 1,847,931 ops/sec ±0.71% (85 runs sampled)
color x 554,733 ops/sec ±0.78% (88 runs sampled)
style x 1,410,772 ops/sec ±0.77% (87 runs sampled)
width array x 189,809 ops/sec ±0.66% (89 runs sampled)
space array x 169,302 ops/sec ±0.62% (89 runs sampled)
fontSize array x 236,943 ops/sec ±0.54% (91 runs sampled)
color array x 200,609 ops/sec ±0.49% (89 runs sampled)
```

*Baseline before this change:*
```
space x 443,754 ops/sec ±0.66% (88 runs sampled)
width x 1,679,140 ops/sec ±0.99% (89 runs sampled)
fontSize x 1,856,617 ops/sec ±0.85% (86 runs sampled)
color x 563,012 ops/sec ±0.87% (88 runs sampled)
style x 1,428,376 ops/sec ±0.74% (88 runs sampled)
width array x 187,758 ops/sec ±0.71% (90 runs sampled)
space array x 156,278 ops/sec ±0.90% (82 runs sampled)
fontSize array x 232,481 ops/sec ±0.54% (86 runs sampled)
color array x 197,372 ops/sec ±0.60% (87 runs sampled)
```

Which suggests the cost when integrated is negligible.

@jxnblk I leave it up to you, but using these benchmarks it seems safe.

aczekajski:
What code did you perform this benchmark on?

If you didn't see, I've created a pull request https://github.com/jxnblk/styled-system/pull/303 with my proposition. I turned out not to be as simple as doing what I've wrote above but still should be pretty fast.

BrendanFDMoore:
@aczekajski sorry for being ambiguous. My previous comment was overall benchmark info on my change for @jxnblk . Unlike my earlier post, this measures the impact on the overall system, rather than the atomic function (which is obviously going to be slower with an extra check)

aczekajski:
Since `px(n)` doesn't cause too much of a performance loss, I won't insist to use my solution. It's up to @jxnblk

aczekajski:
Any updates on this? 😉

jxnblk:
Sorry for the delay!

@BrendanFDMoore thanks for checking the benchmarks on this! Seems like it doesn't appear to have a huge impact, so planning on merging this in and publishing a new version.

@aczekajski planning on going with this one over #303 but thanks for working on this!

PR #297, opened 2018-09-25
getTheme() falsy values

### LINKED ISSUE #254: themeGet() can't return falsy values
I was digging around in [utils.js](https://github.com/jxnblk/styled-system/blob/master/src/util.js#L22-L23) and noticed a bug in (depending on how you look at it) `get()` and/or `themeGet()`: Because both of those functions short-circuit on the value, they will return `null` for falsy values, including `0`:

```js
const theme = {
  borderWidth: [0, 1, 4]
}
get(theme, 'border.0')
themeGet('border.0')({theme})
```

Just to be sure, I made [a test case](https://gist.github.com/shawnbot/ef77715337dd5cadba20c679c130e93b) that fails. It looks like the fix would need to happen in two places:

1. [In `get()`](https://github.com/jxnblk/styled-system/blob/master/src/util.js#L22), change the `(a && a[b])` check to `(a && (b in a))`.
1. [In `themeGet()`](https://github.com/jxnblk/styled-system/blob/master/src/util.js#L25), only return the fallback if `get()` returns `null` or `undefined`.

thompsongl:
Fixes #254 

Uses the prescribed fix from @shawnbot and adds a test

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -21,9 +21,12 @@ export const num = n => typeof n === 'number' && !isNaN(n)
 export const px = n => num(n) ? n + 'px' : n
 
 export const get = (obj, ...paths) => paths.join('.').split('.')
-  .reduce((a, b) => (a && a[b]) ? a[b] : null, obj)
+  .reduce((a, b) => (a && (b in a)) ? a[b] : null, obj)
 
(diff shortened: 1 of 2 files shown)


jxnblk:
Thanks!

PR #312, opened 2018-10-12
Grid gap props should use default spacing scale

### LINKED ISSUE #277: style() function doesn't use default space
All props defined with `style()` function, like `gridGap`, won't use the [default scale](https://github.com/jxnblk/styled-system/blob/aee03959b98022ff7bd41be74ceff36ecae26784/src/space.js#L53-L55).

I'm not sure that `style()` function should do it, but at least props like like `gridGap` should use it for consistency with other props like `m` or `p`.

I'm happy to submit a pull request for that.

sapegin:
Looks like it make sense only for this three props: `gridGap`, `gridColumnGap` and `gridRowGap`.

Fixes #277


src/index.js
```diff
@@ -442,19 +442,22 @@ export const order = style({
 export const gridGap = style({
   prop: 'gridGap',
   transformValue: px,
-  key: 'space'
+  key: 'space',
+  scale: defaultScale
```
flybayer: I don't think you want to use `defaultScale` here because this would always use the default scale even when a different scale is set in the theme, right?

Btw, thanks for the PR to fix this!

jxnblk: Thanks for the PR!

That's right about the `scale` argument – though this is making me wonder whether or not this should be considered a breaking change. If someone is using `gridGap` with a value of, say, `1` then this would cause some layout issues

johno: I think this will be a breaking (but definitely desirable) change.

sapegin: Yeah, looks like it is breaking change even if it won't break any real code ;-)

jxnblk:
Planning on adding this in to v4 so going to go ahead and merge into master. Thanks again for this!

PR #317, opened 2018-10-19
Add cross-env to build on Windows

m0xx:
Couldn''t use package.json build scripts on Windows


jxnblk:
windows doesn't do env vars?! 😧

m0xx:
Yeah but it doesn’t support this syntax :/ that why cross-env is useful. Thx for this project by the way! I tried it yesterday and I it’s awesome!

PR #350, opened 2018-12-16
Update Table of Style Functions to include px, py, mx, my

notgiorgi:
This doc lists `px py mx my` props: https://github.com/jxnblk/styled-system/blob/master/docs/api.md#space

While this doesn't: https://github.com/jxnblk/styled-system/blob/master/docs/table.md


johno:
Thanks!

PR #352, opened 2018-12-21
feat(space): allow nested objects in space

### LINKED ISSUE #351: Inconsistent property access
There's a *big* assumption in the `getValue` function that it operates on a scale (an array of primitive values), it's in line 168:

```js
return px(scale[n] || n)
```

That index/property-access means it'll only take a single-level object, which isn't an assumption i've found in other places. I also don't think it's particularly necessary, as you have a `get` function. This caused me a big headache, as for example `color` behaves differently. I think I can fix this with a backwards-compatible change (simply swapping it property access for `get`), would you be interested in that?

StevenLangbroek:
This uses the internal `get` helper to allow nested property access in `space`, rather than direct. This way space is ambivalent to the data structure of the theme, like other props are.

Are there other props that could use the same treatment?

Fixes #351

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "sideEffects": false,
   "scripts": {
     "prepare": "npm run clean && npm run build:cjs && npm run build:esm",
+    "prettier": "prettier \"{src,test}/**/*.js\"",
     "clean": "rm -rf dist && mkdir dist",
     "build:cjs": "cross-env NODE_ENV=cjs babel src -o dist/index.cjs.js",
     "build:esm": "cross-env NODE_ENV=esm babel src -o dist/index.esm.js",
@@ -41,6 +42,7 @@
     "is-ci": "^1.1.0",
     "mdx-go": "^2.0.0-29",
     "nyc": "^12.0.2",
+    "prettier": "^1.15.3",
     "react": "^16.4.0",
     "react-live": "^1.11.0",
     "react-test-renderer": "^16.4.0",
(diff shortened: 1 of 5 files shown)


StevenLangbroek:
Totally didn't realize this project doesn't have prettier, and I had it enabled in VS Code... Sorry! Want me to add a `.prettierrc`?

jxnblk:
The actual change LGTM – if you want to add a `.prettierrc` file, I think this should be mostly inline with the current code style:

```json
{
  "semi": false,
  "singleQuote": true,
  "trailingComma": "es5"
}
```

StevenLangbroek:
Added `.prettierrc` and `prettier` task to `package.json`. Only goes over `src` and `tests` folders, do you want `packages` as well? If you want I can add a precommit hook as well with `husky`.

StevenLangbroek:
@jxnblk how can I help getting this merged?

jxnblk:
Sorry it's the holidays and I've been away for a bit. Will review this again and hopefully merge soon

package.json
```diff
@@ -7,6 +7,7 @@
   "sideEffects": false,
   "scripts": {
     "prepare": "npm run clean && npm run build:cjs && npm run build:esm",
+    "prettier": "prettier \"{src,test}/**/*.js\"",
```
jxnblk: I've used lint-staged for prettier in other projects. This is fine as-is but might be worth adding that as a follow up

StevenLangbroek: Yeah I'll open a new PR for that once this has been merged.

StevenLangbroek:
Happy new year :)

jxnblk:
Thanks for this! Should be published as v3.1.12 now 🎉

PR #355, opened 2018-12-28
added renderStyledSystemProps helper

matthova:
If I am creating a reusable component with responsive styles, I will want to declare the styles directly inside of the styled component.

For example:
```jsx
const Section = styled.div`
  ${space}
`
```

I can inject responsive styles with an extra function like this:
```jsx
const ResponsiveSection = props => <Section {...props} py={['192px', '224px', '320px']} />;
```

Or I could use `@rebass/components` and do something like this:
```jsx
const Section = system(
  {
    py: ['192px', '224px', '320px'],
  },
  'space',
);
```

Or, per @flybayer's advice, I could use default props like this:
```jsx
const Section = styled.div`
  ${space}
`
Section.defaultProps = {
  py: ['192px', '224px', '320px']
};
```

The goal of this PR is to avoid the workflows described above and instead add a helper function for rendering `styled-system` props inside of a styled component.

Usage of the helper function would look like this:
 ```jsx
 import styled from 'styled-components';
 import { renderStyledSystemProps } from 'styled-system';
 
 const ResponsiveStyledComponent = styled.div`
   ${renderStyledSystemProps({ py: ['192px', '224px', '320px'] })}
 ` 
 ```

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -1292,6 +1292,10 @@ test('renderStyledSystemProps returns an empty object if nothing is passed in', 
   t.deepEqual(renderStyledSystemProps({})({}), {});
 });
 
+test('renderStyledSystemProps returns an empty object if there is no match to a style prop', t => {
+  t.deepEqual(renderStyledSystemProps({ foo: 'bar' })(), {});
+});
+
 test('renderStyledSystemProps converts styled system specific props into generic style props', t => {
   // Test a standard style, "fontSize"
 
--- a/test/index.js
+++ b/test/index.js
@@ -1285,6 +1285,69 @@ Object.keys(styles).forEach(key => {
   })
 })
 
+test('renderStyledSystemProps returns an empty object if nothing is passed in', t => {
+  t.deepEqual(renderStyledSystemProps(  )(  ), {});
+  t.deepEqual(renderStyledSystemProps({})(  ), {});
+  t.deepEqual(renderStyledSystemProps(  )({}), {});
+  t.deepEqual(renderStyledSystemProps({})({}), {});
+});
+
 test('renderStyledSystemProps converts styled system specific props into generic style props', t => {
-  // Need to do something better here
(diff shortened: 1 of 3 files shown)


test/index.js
```diff
@@ -1282,3 +1284,7 @@ Object.keys(styles).forEach(key => {
     })
   })
 })
+
+test('renderStyledSystemProps converts styled system specific props into generic style props', t => {
+  // Need to do something better here
```
matthova: I'm a bit stumped on what would be the best way to test this

matthova: Stumped no more 👍  added a test suite

jxnblk:
Thanks for the PR! One of the goals with v3 was to reduce the size of this library down to the bare minimum and optimize for performance. I'm not sure whether or not the problems this PR attempts to solve are worth including this in the core package – and I think using `defaultProps` is a more React-like (and familiar) API – but going to leave this PR open for discussion for a bit

jxnblk:
Thanks! I don't know if there's anything I've seen *quite* like this approach, but you could always publish your own library 😄

StevenLangbroek:
@matthova we provide an `of` helper for the base components that we use to build more complex components:

```js
// before (the styled-system props comes after the spread
// so that they can't be overridden.
Card.Title = props => (
	<Text
		{...props}
		fontFamily="heading"
		fontSize={1}
		fontWeight="bold"
	/>
);

// after (ordering etc is taken care of)
Card.Title = Text.of({
  fontFamily: 'heading',
  fontSize: 1,
  fontWeight: bold
});
```

jxnblk:
Thanks again for the PR. I'm in the middle of getting v4 together, and I'm not sure that I'd want to introduce this API change into the core library and am closing this for now. That said, I think this could live on as its own utility package built on top of styled-system. Hope that makes sense!

PR #361, opened 2019-01-08
Allow breakpoint array to be used together with breakpoint object

jonespen:
We would really like to use the new breakpoint object syntax (https://github.com/jxnblk/styled-system/pull/341), but currently we have a ton props using the array syntax (`[1, null, null, 2`]). If we could take an incremental upgrade approach, it would make the transition much easier. 

If there is another way doing this, feel free to discard.

Otherwise, mad props for this lib 👍 

cc @WebSeed

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -66,9 +66,17 @@ const getStyles = ({ props, style, value }) => {
   }
 
   // how to hoist this up??
-  const breakpoints = get(props.theme, 'breakpoints') || defaultBreakpoints
+  let breakpoints = get(props.theme, 'breakpoints') || defaultBreakpoints
 
   if (isArray(value)) {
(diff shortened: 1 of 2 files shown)


jxnblk:
Thanks for this! I *think* using aliases (assigning properties to the array) is probably a good way to transition from array syntax to object syntax.

E.g.

```js
const breakpoints = ['40em', '52em', '64em']
breakpoints.small = breakpoints[0]
breakpoints.medium = breakpoints[1]
breakpoints.large = breakpoints[2]
```

I might be missing something, but is there any reason this approach wouldn't work for you?

jonespen:
Oh wow, I completely forgot about the alias syntax (js 🙃), that worked perfectly. Thanks!

PR #377, opened 2019-02-03
Remove unused imports in tests

macklinu:
My IDE highlighted these imports as unused, and when investigating `src/index.js`, I could not find exported values with these names, so I believe these are safe to remove.


jxnblk:
Thanks! I think these were left over from the previous version

PR #383, opened 2019-02-11
V4

jxnblk:
Rewritten core with mostly the same API and similar performance to v3.

- Returns arrays of styles (instead of merging) for responsive and composed styles. This should work the same as before when combined with CSS-in-JS libraries like styled-components and emotion
- Refactors `space` function to use core `style` and `compose` functions
- Reintroduces the `alias` argument for the `style` utility
- Adds long-form props for margin and padding. Shorthand aliases still work.
- The `get` utility works differently, returning the last argument as a fallback.
- Removes `mixed` utility (as it didn't behave as most people would expect)
- The theme scale is now passed as the second argument to the `transformValue` option in `style`
- Removes `ratio` function
- Removes the `styles` and `util` exports
- ~~Removes `meta` field from `propTypes` - this was used by system-docs. An alternative/optional object export for documentation will be added~~
- Removes the `merge` utility
- Changes to border functions:
  - Adds props for `borderStyle`, `borderWidth`, `borderColor`, and `borderRadius`
  - The shorthand border props no longer convert numbers to pixel values with a solid border style
  - All border props are also available as individual exports


package.json
```diff
@@ -74,7 +78,18 @@
     "url": "git+https://github.com/jxnblk/styled-system.git"
   },
   "dependencies": {
-    "@babel/runtime": "^7.1.2",
+    "@babel/runtime": "^7.3.1",
     "prop-types": "^15.6.2"
```
wtgtybhertgeghgtwtg: Probably not the best place to ask this, but have you considered dropping support for `prop-types`?

jxnblk: Seems like an appropriate place to ask. What would the motivation for dropping prop-types be? I suspect these are still useful to people not using TS

flybayer: I prefer not to remove them because I'm currently using them for documenting available props in a [docz](http://docz.site/) site.

wtgtybhertgeghgtwtg: > What would the motivation for dropping prop-types be?

They're no longer coupled with React and, as you said, superfluous to people using (TypeScript or Flow).  It's also pretty easy for a developer to keep them in a production build if they're not using something like `next` or `create-react-app`.

chasemccoy:
@jxnblk Why remove the `styles` export? I have a single file where I take all of the style functions and compose them (using the `compose` util) into groups (layout, typography, etc.) Removing the `styles` export means I will have to refactor to import every single style function individually, which seems messy.

jxnblk:
@chasemccoy it sounds like it would mean defining the object in user code rather than having it in the core library even when not used (by others), but I might misunderstand. I *think* the optional object export for documentation purposes would include keys for all the built-in style functions and could be used for what you're describing though

chasemccoy:
This file from Primer’s component library is almost exactly how we do it:

https://github.com/primer/components/blob/master/src/constants.js

You can see by importing the styles object there it was easy to compose groups of style functions without importing each individually. 

No big deal if this gets removed, just seems to be useful to more than just myself so wanted to bring it up.

jxnblk:
@chasemccoy In the Primer example, using `import * as system from 'styled-system'` could also work, right? The difference would be that the `system` object would also include other utility functions that are exported from this lib

chasemccoy:
@jxnblk Yep, that works!

PR #393, opened 2019-02-21
Add `themeGet` object literal syntax example

fk:
Also adjusts the link to `get` in the main README, and adds `propTypes` there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,7 @@ To learn more, see the [Getting Started](docs/getting-started.md) guide or read 
   - [Variants](docs/api.md#variants)
   - [Utilities](docs/api.md#utilities)
     - [themeGet](docs/api.md#themeget)
+    - [propTypes](docs/api.md#proptypes)
   - [Customize](docs/api.md#customize)
     - [style](docs/api.md#style)
     - [variant](docs/api.md#variant)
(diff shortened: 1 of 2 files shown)


fk:
Meh, reading the updated `themeGet` part on GitHub I have to admit I don't like that too much.
Maybe something like

```js
import { themeGet } from 'styled-system'

const Box = styled.div`
  border-radius: ${themeGet('radii.small', '4px')};
`

// When used with object literal syntax, `themeGet` needs to be in a
// function call and have `props` passed to it:
const AnotherBox = styled('div')(props => ({
  borderRadius: themeGet('radii.small', '4px')(props),
}))
```

jxnblk:
This is great, thanks!

Personally, I like the way it's split up in `api.md` – was there anything in particular you were unsure about? Would love to merge this in, since I think it really clarifies how the utility works

fk:
> was there anything in particular you were unsure about

It just felt a tad too much, and I also started wondering if there are any other object literal examples in the docs at all … probably overthinking ;-)
In this particular case it makes sense to have a bespoke example IMO.

I'm happy with either variant; maybe the more explicit one _is_ indeed better because it is just that! 😅

PR #401, opened 2019-02-26
Add missing getting started link on home page

johno:


johno:
Realized after reading again the getting started content is _on_ the home page 🤦‍♂️

PR #402, opened 2019-02-27
Update README with info on use with object styles

alanbsmith:
I didn't see any existing documentation or conversation about using styled-system with object styles, so I thought I'd add it. 😄 I'm not sure if you'd rather have this in the examples or the docs site instead of the main Readme, but I'm happy to move it wherever. Just let me know.

Thanks for maintaining this lib!

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,37 @@ that handle styles based on values defined in a theme.
 <Box bg='tomato' />
 ```
 
+### Use with Object Styles
+
+*Styled Components*
+
+```jsx
+import styled from 'styled-components';
+import { color } from 'styled-system';
+
+const Text = styled.span(props => (
+  {
+    color: color(props),
+    fontSize: '16px',
+  }
+));
+```
+
+*Emotion*
+
+```jsx
+import styled from '@emotion/styled';
+import { color } from 'styled-system';
+
+const Text = styled.span(
+  {
+    fontSize: '16px'
+  },
+  color
+);
+```
+
+
 ## Responsive Style Props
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -134,6 +134,37 @@ that handle styles based on values defined in a theme.
...
+
+*Styled Components*
+
+```jsx
+import styled from 'styled-components';
+import { color } from 'styled-system';
+
+const Text = styled.span(props => (
+  {
+    color: color(props),
```
alanbsmith: There might be a better way to do this in SC. I was specifically solving for the Emotion use case.

jxnblk: Both styled-components and emotion can accept multiple arguments, so the example you have below for emotion probably makes sense to use as a generic one – I don't think there's a need for both examples

jxnblk:
I suspect the README could use an update in general, but I think this sort of content makes sense in both the readme and `docs/getting-started.md`

MioQuispe:
Here's the cleanest solution I've found for Fela. Do you happen to know if there's a better way? @jxnblk 

`const Box = ({ children, ...props }) => {
  const { css } = useFela();
  return (
    <div className={css(space(props), width(props), fontSize(props), color(props))}>
      {children}
    </div>
  );
};`

README.md
alanbsmith: Cool. I have some time today to update this. Thanks for the feedback!

jxnblk:
@alanbsmith Thanks for putting this together! After a little discussion, I don't think the README is the best place to capture some of this since this is a slightly more advanced usage – we'll definitely be using this to guide how we make the docs better at introducing this – probably in the *Responsive Styles* page

PR #409, opened 2019-03-03
fix: move clean-tag react dep to peers

### LINKED ISSUE #406: React is a direct dependency of clean-tag
https://github.com/styled-system/styled-system/blob/master/packages/clean-tag/package.json#L16

Hello!

clean-tag has React as a direct dependency any chance this could be moved to a peer dep?

It's really only an issue for us when we're linking modules to be honest, albeit annoying.

Cheers

StevenLangbroek:
fixes #406

### DIFF
--- a/packages/clean-tag/package.json
+++ b/packages/clean-tag/package.json
@@ -13,7 +13,6 @@
   "license": "MIT",
   "dependencies": {
     "html-tags": "^2.0.0",
-    "react": "^16.8.1",
     "styled-system": "^3.2.1"
   },
   "devDependencies": {
@@ -30,5 +29,8 @@
       "babel-register"
     ],
     "babel": "inherit"
+  },
+  "peerDependencies": {
+    "react": "^0.14.x || ^15.x || ^16.x"
   }
 }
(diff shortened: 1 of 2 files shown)


StevenLangbroek:
I can't think of a good reason this would be in deps... @jxnblk?

packages/clean-tag/package.json
```diff
@@ -30,5 +29,8 @@
       "babel-register"
     ],
     "babel": "inherit"
+  },
+  "peerDependencies": {
+    "react": "^0.14.x || ^15.x || ^16.x"
```
jxnblk: I think `React.forwardRef` requires 16.x (can't remember which minor version): https://github.com/styled-system/styled-system/blob/master/packages/clean-tag/index.js#L20

MatthewLarner: Thanks for looking into this!

https://github.com/facebook/react/blob/master/CHANGELOG.md#1630-march-29-2018

16.3 for `forwardRef `

jxnblk:
Closing this out in favor of #433 – this issue can be addressed in the new repo if needed

PR #410, opened 2019-03-03
feat: rename blacklist to omitProps

### LINKED ISSUE #391: Discussion: blacklist / whitelist terminology vs inclusiveness
I'd like to bring up that there are issues with the term blacklist. A recent incident:

https://twitter.com/BenLesh/status/1095487146251546624

Since you're close to releasing a breaking version, maybe this is a good time to consider changing this terminology to something more inclusive? I'd be more than willing to make the changes to code & documentation if you agree this is the right thing to do. If you think it is, but not now, I'd love to hear when you think is :).

Thanks ❤️

StevenLangbroek:
fixes #391

BREAKING CHANGE: `blacklist` has been renamed to `omitProps`, so consumers
will need to change this.

### DIFF
--- a/packages/clean-tag/README.md
+++ b/packages/clean-tag/README.md
@@ -45,15 +45,15 @@ Changes the underlying HTML element per component instance.
 <Heading is='h1' />
 ```
 
-## `blacklist` prop
+## `omitProps` prop
 
-Provide a custom blacklist of prop names.
+Provide a custom list of prop keys to remove from the backing styled component.
 
-The default blacklist is based on Styled System's propType definitions.
+The default list is based on Styled System's propType definitions.
 
 ```js
 <Heading
-  blacklist={[
+  omitProps={[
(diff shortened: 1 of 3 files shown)


StevenLangbroek:
@jxnblk I found the terminology in some internal stuff as well:

image

You mentioned `system-components` was superseded by `@rebass/components`, should I just leave it for deprecation? It's also in `system-classnames`, I can either do it as part of this PR or open a separate one 🤷‍♂️.

jxnblk:
Thanks! Yeah I'll open another PR to remove system-components, but renaming the variable in `system-classnames` here would be helpful – also should ditch that language in `@rebass/components` at some point

jxnblk:
The deprecated `system-components` package was removed in #422

jxnblk:
Thanks for this! Going to go ahead and merge this in

Issue #375, opened 2019-01-31
Add website icon 


mukeshmandiwal:
Currently, the website doesn't have any favicon in the header section.

We need to add an icon for the website. We can use existing `logo.png` from docs folder for this.

jxnblk:
I'm in the process of completely redoing the docs, so if you want to hold off until that's ready that'd be helpful

mukeshmandiwal:
yea sure

mukeshmandiwal:
I am not able to add the website icon because public folder added in  .gitignore

jxnblk:
For Gatsby sites, there's a `static/` folder, see https://www.gatsbyjs.org/docs/static-folder/

mfix22:
This can be closed @jxnblk 👍

PR #412, opened 2019-03-04
Add website icon

mukeshmandiwal:


mukeshmandiwal:
fixes #375

crashuniverse:
We can remove the 2nd commit that adds delectation message by rebasing with latest branch. The PR is easier to focus if it has only your contribution.

docs/src/Root.js
```diff
@@ -34,6 +34,7 @@ const Root = props => {
     <>
       <Helmet>
         <title>{title}</title>
+        <link rel="shortcut icon" type="image/png" href="/logo.png"/>
```
crashuniverse: Logo in this case is 512 size, maybe a smaller 64x64 file for favicon.png or favigon.ico would be a better usage than logo.

karlhorky:
Maybe we should have a background color for dark mode? So that the favicon doesn't show black on dark gray like GitHub?

Screen Shot 2019-04-09 at 11 17 07

mukeshmandiwal:
Should we use logo-white.png as favicon ?

karlhorky:
Hm, that will have problems on non-dark mode tabs, wouldn't it?

I guess my suggestion would be to fill in the transparent parts as well as a 2 pixel margin around it with white (2 pixels when it's shrunk to favicon size).

Something that would look like this:

karlhorky:
You may also want to consider generating some other versions of the favicon with something like RealFaviconGenerator (you could take a look at my PR here: keepassxreboot/keepassxreboot.github.io#54)

docs/src/index.js
```diff
@@ -134,6 +134,7 @@ const Page = props => {
     <>
       <Helmet>
         <title>{title}</title>
+        <link rel='shortcut icon' type='image/x-icon' href='/favicon.ico'/>
```
jxnblk: Thanks for looking into this! Whatever you end up using to generate the icon with, can you please add an `npx` script to the `package.json`? – see how repng is used for the other images

mukeshmandiwal: I use [https://realfavicongenerator.net/](url) link to generate favicon from static/logo.svg file

jxnblk: Thanks! If there’s not an installable package this will likely fall out of sync over time

mukeshmandiwal: As of my concern, there is no installable package which likely fulfills our requirements

karlhorky: Maybe one of the following npm packages? I just did a quick search for "svg to ico npm" on Google:

- https://www.npmjs.com/package/favicons
- https://www.npmjs.com/package/icon-gen

jxnblk: It could also be generated from the `png` output from repng, if that makes finding a library easier

jxnblk: Is there any reason *not* to use a `.png` file for the favicon? I assume most modern browsers support this now, but haven't done any research

jxnblk:
Thanks again for the work on this! I'll go ahead and merge this PR since I'm getting ready to ship updated docs for v5

PR #417, opened 2019-03-07
Allow theme.space to be array *or* object

jaridmargolin:
This functionality appears to have been removed from v4 (hopefully not intentionally). Currently all other theme props support being defined as an array or an object. Spacing properties are the exception (or the only exception I have run into).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -91,14 +91,14 @@ export const style = ({
   }
   func.propTypes[prop].meta = {
     prop,
-    themeKey: key
+    themeKey: key,
   }
 
   if (alias) {
     func.propTypes[alias] = cloneFunction(propType)
     func.propTypes[alias].meta = {
       prop: alias,
-      themeKey: key
+      themeKey: key,
     }
   }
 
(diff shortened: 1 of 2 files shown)


jaridmargolin:
I've also hit a few occasions where it would be helpful to declare negative margins/paddings. I was considering adding support for this by checking if object key startsWith `-`. but was concerned this would be unexpected behavior w/ objects. Any thoughts?

aholachek:
Merging this pr would be helpful for my use case &mdash; designers have passed down a series of named spacing increments to integrate into our design system, and I'd prefer to stay close to their naming scheme rather than introducing a new mental model (array indices) just for spacing.

jxnblk:
Thanks! Yeah, this was an unintentional regression so thanks for catching this!

jaridmargolin:
@jxnblk - Thank you for the quick merge/release. Do you have any opinions on if object definitions should support negative values? Ex: `mx='-small'`

I would have to sniff for `-` as the first character, which feels a little hacky. At the same time, it's annoying to NOT have access to negative theme values.

PR #429, opened 2019-03-20
Fix border props sequence

kkir:
Fixes following issue:

let's say we have a component:
```js
const Component = () => styled.div`
  ${borders}
`
```
then we render it with the following props:
```js
<Component borderBottom="1px solid" borderColor="red" />
```
than CSS will be:
```css
border-color: red;
border-bottom: 1px solid;
```
the `red` color will be replaced with `inherit` which is `#000` in Chrome 72

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -471,11 +471,11 @@ export const borders = compose(
   border,
   borderStyle,
   borderWidth,
-  borderColor,
   borderTop,
   borderRight,
   borderBottom,
   borderLeft,
+  borderColor,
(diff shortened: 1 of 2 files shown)


jxnblk:
Good catch, thanks! Should `borderStyle` and `borderWidth` be rearranged as well?

kkir:
@jxnblk absolutely!
Thanks for the awesome lib!

jxnblk:
Thanks for this! Going to merge and :shipit:

PR #432, opened 2019-03-22
Themes spec ideas

giuseppeg:
👋 Brent, I love the idea of a standard theming format as it is something that I've been wanting for a long while - specifically since the day Jina and some other salesforce folks showed us (yelp - my former employer) this concept of design tokens. Since then I tried to use [a similar concept](https://github.com/giuseppeg/suitcss-toolkit/tree/example-app/examples/app#themes) in suit css (i am one of the maintainers although the activity on that project is very limited nowadays) and [personal projects of mine](https://twitter.com/giuseppegurgone/status/1019431376020561921).

I made some changes (ideas/suggestions) to the spec mainly to start off a discussion/brainstorming around those. The PR is not intended to be merged and therefore not polished :)

### DIFF
--- a/docs/theme-specification.md
+++ b/docs/theme-specification.md
@@ -7,49 +7,88 @@ The object itself is not coupled to Styled System's implementation and can be us
 ## Scale Objects
 
 Many CSS style properties accept open-ended values like lengths, colors, and font names.
-In order to create a consistent styling system, the theme object is centered around the idea of scales, such as a typographic (font-size) scale, a spacing scale for margin and padding, and a color object.
-These scales can be defined in multiple ways depending on needs, but tend to use arrays for ordinal values like font sizes, or plain objects for named values like colors, with the option of nesting objects for more complex systems.
+In order to create a consistent styling system, the theme object is centered around the idea of scales, such as a typographic (font-size) scale, a spacing scale for margin and padding, and a color object. Scales must be JSON serializable - this way can be used in JS and CSS but possibly even cross platform.
+
+These scales can be defined after their correspective CSS property names when possible and use arrays for variants or ordinal values like font sizes. Short hand values are expressed as nested objects. Special categories can be defined to describe multiple CSS properties whichmust be listed in the `aliases` part of the object.
+
+Each scale must contain exactly 7 variants, where the first one is the base value.
+(not sure what is the best way to define base + variants values though - maybe your idea of singular and plural is much better).
+
+The idea is to mirror the HTML heading scale h1-6 + a base value - 6 variants are normally good enough. I might be wrong though, is 10 or Pantone-style better?
 
 ```js
-// example fontSizes scale as an array
-fontSizes: [
-  12, 14, 16, 20, 24, 32
-]
-```
-
-```js
-// example colors object
-colors: {
-  blue: '#07c',
-  green: '#0fa',
-}
-```
-
-```js
-// example nested colors object
-colors: {
-  blue: '#07c',
-  blues: [
-    '#004170',
-    '#006fbe',
-    '#2d8fd5',
-    '#5aa7de',
+// example font scale as an array
+font: {
+  size: [ // font-size or fontSize
+    16, // base
+    12, 14, 18, 20, 24, 32
+  ],
+  family: {
+    sansSerif: 'Open Sans, Arial',
+    serif: 'times',
+    monospace: 'monospace'
+  },
+  weight: [
+    300, // base
+    100, 200, 400, 500, 600, 700
   ]
 }
 ```
 
-### Scale Aliases
+```js
+color: {
+  blue: [
+    '#07c', // base
+    '#004170',
+    '#006fbe',
+    '#2d8fd5',
+    '#5aa7de',
+    '#aaa',
+    '#bbb',
+    '#ccc'
+  ]
+}
+```
 
-For typically ordinal values like font sizes that are stored in arrays, it can be helpful to create aliases by adding named properties to the object.
+### ~~Scale Aliases~~
+
+~~For typically ordinal values like font sizes that are stored in arrays, it can be helpful to create aliases by adding named properties to the object.~~
+
+I think that these kind of aliases must be done in user land with ad-hoc maps otherwise they won't be universal or portable e.g. i create a `foo` alias and use it in my codebase, then I import your theme which doesn't define it or calls it `bar` and things won't work.
+
+See https://github.com/giuseppeg/dss/blob/master/website/theme/index.js for an example of map.
+
+Instead aliases should be for things like `space` (probably automated) and only top level keys:
 
 ```js
-// example fontSizes aliases
-fontSizes: [
-  12, 14, 16, 20, 24, 32
-]
-// aliases
-fontSizes.body = fontSizes[2]
-fontSizes.display = fontSizes[5]
+{
+  font: {
+    size: [ // font-size or fontSize
+      16, // base
+      12, 14, 18, 20, 24, 32
+    ]
+  },
+  space: [
+    2, 4, 8, 16, 32, 64
+  ],
+  aliases: {
+    margin: 'space',
+    padding: 'space',
+  }
+}
+```
+So that a tool can do:
+
+```js
+function propName(prop) {
+  return theme.aliases[prop] && theme[theme.aliases[prop]] || theme[prop]
+}
+
+<Box margin="0" />
+
+// vvvv
+
+propName('margin')[0]
 ```
 
 ### Excluded Values
@@ -61,8 +100,8 @@ Other properties that are intentionally excluded from this specification include
 
 ## Keys
 
-The keys in the theme object should typically correspond with the CSS properties they are used for, and follow a plural naming convention.
-For example, the CSS property `font-size` is expected to use values from the `fontSizes` scale, and the `color` property uses values from the `colors` scale.
+The keys in the theme object should typically correspond with the CSS properties they are used for.
+For example, the CSS property `font-size` is expected to use values from the `font.size` scale, and the `color` property uses values from the `color.someColorName` scale.
 
 Some keys can be used for multiple CSS properties, where the data type is the same. The `color` object is intended to be used with any property that accepts a CSS color value, such as `background-color` or `border-color`.
 
@@ -70,35 +109,25 @@ Some keys can be used for multiple CSS properties, where the data type is the sa
 ### Space
 
 The `space` key is a specially-named scale intended for use with margin, padding, and other layout-related CSS properties.
-A space scale can be defined as either a plain object or an array, but by convention an array is preferred.
+A space scale must be defined as an array.
 This is an intentional constraint that makes it difficult to add *"one-off"* or *"in-between"* sizes that could lead to unwanted and rippling affects to layout.
 
 When defining space scales as an array, it is conventional to use the value `0` as the first value so that `space[0] === 0`.
 
 ```js
-// example space scale
+// GOOD: example space scale
 space: [
   0, 4, 8, 16, 32, 64
 ]
 ```
 
 ```js
-// example space scale object
+// BAD not standard and opinionated naming
 space: {
   small: 4,
   medium: 8,
   large: 16,
 }
-```
-
-```js
-// example space scale with aliases
-space: [
-  0, 4, 8, 16, 32
-]
-space.small = space[1]
-space.medium = space[2]
-space.large = space[3]
 ```
 
 ### Breakpoints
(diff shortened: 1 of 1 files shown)


docs/theme-specification.md
```diff
@@ -7,49 +7,88 @@ The object itself is not coupled to Styled System's implementation and can be us
...
-fontSizes: [
-  12, 14, 16, 20, 24, 32
-]
-```
+These scales can be defined after their correspective CSS property names when possible and use arrays for variants or ordinal values like font sizes. Short hand values are expressed as nested objects. Special categories can be defined to describe multiple CSS properties whichmust be listed in the `aliases` part of the object.
+
+Each scale must contain exactly 7 variants, where the first one is the base value.
+(not sure what is the best way to define base + variants values though - maybe your idea of singular and plural is much better).
+
+The idea is to mirror the HTML heading scale h1-6 + a base value - 6 variants are normally good enough. I might be wrong though, is 10 or Pantone-style better?
```
giuseppeg: specifically for color maybe the human readable keys should be the standard color keywords https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#Color_keywords - the color doesn't have to match the standard hex value and be slightly different but the keyword (property name) would at least be ±universal

jxnblk:
This is awesome, thanks! 🙌  (Huge fan of SuitCSS if that wasn't obvious) There's a lot of good thoughts and starting points for discussion, so I'll try to share my thoughts...

Generally as a first pass at this sort of thing, I don't want to be too restrictive as to hinder adoption. Although I like guidelines or conventions around *what* goes in each scale, I tend to think that's better left open-ended. I.e. one team might want 6 font sizes, one team might want 32, and then someone else might want 2, and I wouldn't want this spec to immediately appear like it doesn't cater to those use-cases. For colors, especially, I feel like there are far too many variations to account for. There are also a fair number of people who are *against* the use of arrays when they first see scales like these, and I'd rather this feel more inclusive overall. That said, perhaps there could be different *levels* of conformance to the spec, where the things you're describing are more like an optional *strict-mode* – curious what you're thoughts are here...

For the scale names themselves, I personally think pluralization makes more sense and is more inline with common naming conventions. E.g. `fontSizes` includes multiple font-sizes, not just one. 

Your point about aliases is a good one, and I'd like to think more on it a bit. Based on issues and questions I've received in this library, allowing the scales to be open-ended plain JS objects helps take care of a lot of edge cases.

I'll add some more comments inline, but really value the perspective here and want to work together to figure out the best common ground for the entire community

docs/theme-specification.md
```diff
@@ -7,49 +7,88 @@ The object itself is not coupled to Styled System's implementation and can be us
 ## Scale Objects
 
 Many CSS style properties accept open-ended values like lengths, colors, and font names.
-In order to create a consistent styling system, the theme object is centered around the idea of scales, such as a typographic (font-size) scale, a spacing scale for margin and padding, and a color object.
-These scales can be defined in multiple ways depending on needs, but tend to use arrays for ordinal values like font sizes, or plain objects for named values like colors, with the option of nesting objects for more complex systems.
+In order to create a consistent styling system, the theme object is centered around the idea of scales, such as a typographic (font-size) scale, a spacing scale for margin and padding, and a color object. Scales must be JSON serializable - this way can be used in JS and CSS but possibly even cross platform.
```
jxnblk: I really like the idea of being JSON serializable – is this also part of the reason you changed the approach to aliasing?

docs/theme-specification.md
```diff
@@ -131,68 +160,55 @@ m: [ 0, 1, 2 ]
 }
 ```
 
-#### Media Queries
-
-For convenience and for use with other styling approaches, a `mediaQueries` scale derived from the `breakpoints` scale can be added to the theme object.
```
jxnblk: Why ditch this? I prefer min-width media queries, but do see the value in allowing other media queries, even `@media print`

docs/theme-specification.md
```diff
@@ -131,68 +160,55 @@ m: [ 0, 1, 2 ]
...
-`maxWidths`       | `max-width`
-`minWidths`       | `min-widths`
-`widths`          | `width`
-`maxHeights`      | `max-height`
-`minHeights`      | `min-height`
-`heights`         | `height`
-`borders`         | `border`, `border-top`, `border-right`, `border-bottom`, `border-left`
-`borderWidths`    | `border-width`
-`borderStyles`    | `border-style`
+`font.size`       | `font-size`
```
jxnblk: I get the logic with nested objects, but also tend to favor flatter structures in cases like this to avoid what I call "premature organization" – that said this is directly from CSS itself, so it could make sense

docs/theme-specification.md
```diff
@@ -131,68 +160,55 @@ m: [ 0, 1, 2 ]
...
+`width`           | `width`
+`maxHeight`       | `max-height`
+`minHeight `      | `min-height`
+`height`          | `height`
+`border`          | `border`, `border-top`, `border-right`, `border-bottom`, `border-left`
+`border.width`    | `border-width`
 `radii`           | `border-radius`
-`shadows`         | `box-shadow`, `text-shadow`
-`zIndices`        | `z-index`
+`boxShadow`       | `box-shadow`, `text-shadow`
```
jxnblk: Would you propose a separate `textShadow` field with this direction?

docs/theme-specification.md
giuseppeg: yeah, I was trying to also to come up with a way to make it easy for tools to make look ups

docs/theme-specification.md
giuseppeg: sorry I went I see small, medium large xxl etc I short circuit and reject the idea :D mostly because there will always be that one-off where maybe you are going to need `aBitBiggerThanSmallButNotMedium`. Also I think that the future is about element queries / resize observer / conditional rendering based on local context rather than global (the entire viewport).

That said the `mediaQueries` section can be useful (and many will keep using media queries) so I'd bring it back. I wonder if there is a better way to classify them that is as idiomatic as `small`,`medium`.

docs/theme-specification.md
giuseppeg: that'd make sense if we go for the singular form and css properties names. Otherwise I think the your original idea is great:

giuseppeg:
Thank you for taking the time to reply! 

I think that I might have missed the point of this spec actually. Probably this is meant to describe a way for tools to support a "standard" theming format (design tokens) rather than a generic way to define themes that people can then swap? 

Basically themes can be very opinionated but follow some basic rules that allow tools (css in js libs for example) to understand them. If that's the case then I totally agree with the following 

> Generally as a first pass at this sort of thing, I don't want to be too restrictive as to hinder adoption. Although I like guidelines or conventions around what goes in each scale, I tend to think that's better left open-ended.

I also agree with the following sentiment

> For colors, especially, I feel like there are far too many variations to account for. There are also a fair number of people who are against the use of arrays when they first see scales like these, and I'd rather this feel more inclusive overall. 

In fact at work right now we use simple variables like so:

```
--color-blue: #07c; /* base */

/* variants */
--color-blue1
--color-blue2
--color-blue3
--color-blue4
--color-blue5
--color-blue6
/* lol */
```

> For the scale names themselves, I personally think pluralization makes more sense and is more inline with common naming conventions

> I get the logic with nested objects, but also tend to favor flatter structures

That works too. So you would use singular for the base value and plurals for variants? Eg.

```js
{
  colors: {
    blue: '#07c',
    blues: [ ... ]
  }
}
```

I suggested to use singular form, nested objects and matching CSS property names because that makes it way easier to write an helper to do lookups eg:

```js
{ 
  font: {
    size: []
  }
}

'font-size'.split('-')// .. and so on
'fontSize'.replace(/[A-Z]/g, '-$&').split('-')// .. and so on
```

docs/theme-specification.md
jxnblk: Yeah, makes sense. Using arrays still leaves room for that "in-between" size issue. I mostly used `small`, `medium`, `large` to convey they idea of what `mediaQueries` was for, but it's certainly not the best naming convention

johno: I think a theme specification should include some notion of breakpoints since it's intimately related to scales, colors, etc. when building/designing. IMO if a collection of media queries can handle 90%+ of the usage then it's a success. There will always be the need for one off mqs where something needs to be in between two values. That's something we don't necessarily need to accommodate.

docs/theme-specification.md
```diff
@@ -131,68 +160,55 @@ m: [ 0, 1, 2 ]
...
+`lineHeight`      | `line-height`
+`color`           | `color`, `background-color`, `border-color`
+`letterSpacing`   | `letter-spacing`
+`maxWidth`        | `max-width`
+`minWidth`        | `min-widths`
+`width`           | `width`
+`maxHeight`       | `max-height`
+`minHeight `      | `min-height`
+`height`          | `height`
+`border`          | `border`, `border-top`, `border-right`, `border-bottom`, `border-left`
```
johno: Wouldn't this conflict with `border.width`?

docs/theme-specification.md
johno: Something also to consider is that if `font` is omitted entirely from a theme, tooling now needs to perform existence checks.

johno: Additionally, "merging" themes in tooling becomes more complex.

docs/theme-specification.md
johno: I personally prefer the plural form since they often map to a collection of values, but it's definitely not a hill I would die on so long as we're consistent.

docs/theme-specification.md
johno: Yeah, ensuring that scales are JSON serializable is 💯. Love that idea.

docs/theme-specification.md
```diff
@@ -61,46 +100,36 @@ Other properties that are intentionally excluded from this specification include
...
+For example, the CSS property `font-size` is expected to use values from the `font.size` scale, and the `color` property uses values from the `color.someColorName` scale.
 
 Some keys can be used for multiple CSS properties, where the data type is the same. The `color` object is intended to be used with any property that accepts a CSS color value, such as `background-color` or `border-color`.
 
 
 ### Space
 
 The `space` key is a specially-named scale intended for use with margin, padding, and other layout-related CSS properties.
-A space scale can be defined as either a plain object or an array, but by convention an array is preferred.
+A space scale must be defined as an array.
```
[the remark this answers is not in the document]
jaridmargolin: We have had serious issues with array based scales and I definitely would not like to see this enforced (for all values, spacing included). We have moved 100% away from array based scales.

* If you do NEED to introduce an additional spacing, it is an absolute nightmare to update your application. `styled-system` can useful for mature design systems, but I have used it heavily for evolving design systems (it is a great way to start enforcing some structure).
* The mental mapping of indexes to values it a deterrent for some developers.

jxnblk:
Not to detract from this PR, but heads up that I've started a new org & repo here: https://github.com/system-ui/theme-specification/ and started an [issue for further discussion](https://github.com/system-ui/theme-specification/issues/1)

giuseppeg:
301 to https://github.com/system-ui/theme-specification/issues/1

docs/theme-specification.md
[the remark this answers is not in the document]
giuseppeg: @jaridmargolin how do you define scales then? Do you use names like small, medium etc?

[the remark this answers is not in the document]
jaridmargolin: We are just keying the values:

```
space: {
  's4': 4,
  's8': 8,
  's12': 12
}
```

Additionally, we haven't been able to determine what wins we get from letting style-system map to theme values (other than that it is slightly less characters to type). We have however determined that we lose type safety. As a result, we are moving away from styled-system resolving theme values.

Our final usage looks like:

```
import theme from 'PATH/TO/THEME'

<Box pl={theme.space.s8}>...</Box>
```

While slightly more verbose this solves both problems I had outlined above (and also gives us some type safety).

---

I don't expect our solution to be utilized by all, but I think the system should be flexible enough to accommodate our use case.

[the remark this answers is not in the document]
giuseppeg: I think you can still work with arrays and get type safety https://twitter.com/giuseppegurgone/status/1084753366176223232 (those could be generated)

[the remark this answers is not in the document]
giuseppeg: I personally don't like the abbreviations/aliases React props as it is a totally new and arbitrary vocabulary to learn. I am more of a fan of a single `style` prop like in React Native, but that's personal preference and doesn't relate to the topic of this PR (themes)

PR #442, opened 2019-03-27
Fix createGlobalStyle in Basic Example

jayhxmo:
createGlobalStyle should not be invoked as a function

Just a beginner here — I noticed that createGlobalStyle does not work as a function call in Basic Example. Was this a mistake?


jxnblk:
Looks like that was a mistake. Thanks!

PR #443, opened 2019-03-30
Docs site update

jxnblk:


joe-bell:
@jxnblk this reference table is so much cleaner, thank you

PR #457, opened 2019-04-13
Optimize space mapProps function

sbrichardson:
The logic with space had multiple unnecessary function calls and duplicated logic. Since it's used a lot within the system, I removed the extra function calls to is() and reduced it's work.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -235,17 +235,14 @@ export const paddingRight = style({
   scale: spaceScale,
 })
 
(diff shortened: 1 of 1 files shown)


jxnblk:
Thanks for this! In light of the changes in #479, I suspect the `mapProps` function will be going away, so I'm closing this one out

sbrichardson:
No problem!

PR #473, opened 2019-04-30
Merge styles into single object

### LINKED ISSUE #471: First class support of styled-system by JSS
Hey, as discussed documenting issues I stumbled over while trying to integrate with JSS:

1. Styled-system functions return value data structure.

    Currently, v4 returns a nested array, which could be a flat object as it has been before in v3.  Seems like @jxnblk and I both want it flattened again 😂

2. Dependency on React prop-types isn't ideal, because it could be done in a universal way. Also it was a bit confusing to find `propTypes` property on those functions, because it made them look like a React component. I am looking to have a first-class support for styled-system and keen to integrate the validation automatically.

Related issue on JSS repo https://github.com/cssinjs/jss/issues/974

jxnblk:
This change undoes the change in v4.0.0 where the returned value is an array of styles, favoring a merged object instead. The merge implementation is loosely based on the `deepmerge` package, but only handles objects and does not check for null, falsy or array values. This shouldn't entail a breaking change for most users, but does change the returned values

Partially fixes #471

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -51,7 +51,13 @@ export const merge = (a, b) => {
   return result
 }
 
-const mergeAll = (...args) => args.reduce((acc, arg) => merge(acc, arg), {})
+const mergeAll = (...args) => {
+  let result = {}
+  for (let i = 0; i < args.length; i++) {
+    result = merge(result, args[i])
+  }
+  return result
+}
 
 export const style = ({
   prop,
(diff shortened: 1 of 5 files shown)


src/index.js
```diff
@@ -35,6 +35,24 @@ export const createMediaQuery = n => `@media screen and (min-width: ${px(n)})`
 
 const getValue = (n, scale) => get(scale, n)
 
+// loosely based on deepmerge package
+export const merge = (a, b) => {
+  const result = {}
+  Object.keys(a).forEach(key => {
```
kof: It might be worth doing micro perf optimizations here by using a for loop instead of .keys, .forEach, .reduce, since this is something that will be done a lot.

PR #479, opened 2019-05-05
Improve performances

neoziro:
Following #444, I bring perf improvements from @smooth-ui/system into styled-system.
The goal is to make this library the best one and the universal one.

What have changed?

- I switched to the core of @smooth-ui/system, more performant
especially with multiple compose. Composition is now flat, it means you
can compose 10 times and it will not be slower. We are probably in O(n),
where "n" is the number of props.
- I switched to rollup in order to optimize build, provide umd and
improve tree shaking

What are the breaking changes?

- We now rely on the order of props. ES2015 is here and we can rely on
it, see this tweet for context: https://twitter.com/ianstormtaylor/status/1115680037875752963.
- Object keys or array keys have now equivalent parsing

See #470 for all history and original code.


neoziro:
@jxnblk are we ready to merge?

src/index.js
```diff
@@ -154,109 +265,113 @@ export const mapProps = mapper => func => {
...
 
+// Space
 export const margin = style({
   prop: 'margin',
   alias: 'm',
+  cssProperties: ['margin'],
   key: 'space',
-  transformValue: getSpace,
-  scale: spaceScale,
+  transformValue: scale(),
```
jxnblk: This doesn't need to be a function call

src/index.js
```diff
@@ -154,109 +265,113 @@ export const mapProps = mapper => func => {
...
-  const absolute = Math.abs(n)
-  const value = get(scale, absolute)
+const scale = (defaultVariants = DEFAULT_SPACE) => (
+  transformedValue,
+  { rawValue, variants = defaultVariants }
+) => {
+  if (!num(rawValue)) {
+    return px(variants[rawValue] || rawValue)
+  }
+  const abs = Math.abs(rawValue)
```
jxnblk: This logic is only needed for negative margins – I'd avoid doing this for scales like `fontSizes`

src/index.js
```diff
@@ -398,39 +487,39 @@ export const size = mapProps(props => ({
 
 export const verticalAlign = style({ prop: 'verticalAlign' })
 
-// flexbox
+// Flexboxes
```
jxnblk:
```suggestion
// flexbox
```

src/index.js
```diff
@@ -398,39 +487,39 @@ export const size = mapProps(props => ({
...
+  transformValue: getWidth,
+})
 export const flexDirection = style({ prop: 'flexDirection' })
 export const flex = style({ prop: 'flex' })
 export const justifySelf = style({ prop: 'justifySelf' })
 export const alignSelf = style({ prop: 'alignSelf' })
 export const order = style({ prop: 'order' })
 
-// grid
+// Grid
```
jxnblk:
```suggestion
// grid
```

src/index.js
```diff
@@ -503,30 +607,30 @@ export const borders = compose(
...
-export const right = style({ prop: 'right', transformValue: getPx })
-export const bottom = style({ prop: 'bottom', transformValue: getPx })
-export const left = style({ prop: 'left', transformValue: getPx })
+export const boxShadow = style({
+  prop: 'boxShadow',
+  key: 'shadows',
+})
 
-// variants
+// Variants
```
jxnblk:
```suggestion
// variants
```

.npmignore
```diff
@@ -1,14 +1,3 @@
...
-docs
-src
-examples
-babel.config.js
-test
-CONTRIBUTING.md
-CODE_OF_CONDUCT.md
+/*
+!/dist/**/*.js
+!/props.js
```
jxnblk: nb: personally I prefer not using negation for the ignore files

benchmarks/v4.1.0.js
```diff
@@ -0,0 +1,688 @@
+"use strict";
```
jxnblk: Since this is copypasta dist code from the current version, I don't think this should be checked in

src/index.js
```diff
@@ -23,127 +41,220 @@ export const get = (obj, ...paths) => {
...
-  for (const key in a) {
-    result[key] = a[key]
+function getBreakpoints(theme) {
+  const themeBreakpoints = getThemeValue(theme, 'breakpoints')
+  const breakpoints = is(themeBreakpoints)
+    ? themeBreakpoints
+    : defaultBreakpoints
+  // TODO deprecate in v5
+  if (Array.isArray(breakpoints)) {
+    const clonedBreakpoints = merge([], breakpoints)
```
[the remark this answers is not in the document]
neoziro: It doesn't work with mixed array and object values. The spread operator of array only spread the iterator, so we miss the properties. `merge` take care of both.

test/index.js
```diff
@@ -117,16 +117,32 @@ test('skips undefined responsive values', t => {
...
+// })
+
+// Object parsing work as expected...
+// Previous behaviour was kind of buggy for me
 test('parses object values', t => {
   const style = width({
     width: {
-      _: '100%',
-      2: '50%',
+      0: '100%',
```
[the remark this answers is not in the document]
neoziro: In fact, the previous behaviour was not correct for me. An object was like an array:

```js
// previous behaviour
{ _: '100%', 2: '50%' } // ['100%', '50%']
// current behaviour
{ _: '100%', 2: '50%' } // [, , '50%']
```

For me it is the correct and the expected behaviour, it is more a bug fix than a breaking change.

test/index.js
```diff
@@ -117,16 +117,32 @@ test('skips undefined responsive values', t => {
...
+      1: '50%',
```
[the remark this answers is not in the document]
neoziro: See previous comment.

neoziro:
Hello @jxnblk, I made the changes and I rebased. One test is not passing, the merge of variant, it was not part of the library initially. Also using your merge method instead of the deepmerge package makes a significant less of perf (according to benchmark).

---

At start, when I come with this PR, I wanted to migrate Smooth UI System to styled-system. But I think it would not be possible, even if the philosophy is very close, we are not going in the same direction. There is a lot of points where we disagree:

- Jest VS ava
- Rollup vs Microbundle
- Split code in several files (code readability)
- Use of external packages (why copying deepmerge instead of including it?)
- References in theme (it costs nothing to support it)
- Naming of properties (color vs textColor?)
- Breakpoints array (why not specifying 0?)
- ...

Also, performance seems to not be something very important for you since a lot of other stuff was implemented and merged before this PR. I try to keep up with rebase but it is difficult (eg. the compose of variant).

I understand your choices, but there are different from mines. So I think we can share the same "theming specification" but I will not use styled-system in Smooth UI or in my projects. I use it professionally, and I have my own roadmap, I can't stick to yours sorry.

Thank you for your time, I hope my contribution helped you!

jxnblk:
@neoziro I'd really love to figure out how to get this landed, and I'd appreciate some patience here. This is a substantial PR and like any large code change, I expect it to take longer than some of the other smaller PRs that have been merged in. I can chip in to help with parts that you aren't interested in working on, but it sounds like maybe this PR's base should change to a `next` branch so that this can land as a major version change instead of in v4.

As for the requested changes, you originally proposed a performance refactor, but also chose to include changes to the build and file structure – and are proposing changing the testing library as well. I'm not opposed to these changes, but they weren't discussed beforehand and make this PR harder to evaluate. I need to make sure changes in v4 do not break for the current users of this library, and I hope you can understand where I'm coming from here

neoziro:
@jxnblk yeah sorry for including these changes, it was not part of perf issue. I think it is going to slow for me, anyway I could continue to submit issues and changes. About the PR, I don’t have time to invest more on it, but the core is here, feel free to iterate on it! Thanks for your time and your patience 🙏

jxnblk:
Going to close this out since it seems like it's not moving forward. Thanks again for all the work on this and sorry this PR didn't work out. Hopefully we can collaborate in the future, and I'll definitely be taking some inspiration from this for the next version of styled-system

neoziro:
@jxnblk no problem. You have the PR for history with the flattening of props in compose. It is the key for keep perf constants with several composes. Always happy to help! Thanks for this great library!

PR #491, opened 2019-05-15
Add the responsive default alias `_` to documentation

grrowl:
I had to dig into the GitHub to find how to apply a default style (my responsive elements were appearing on `< small` breakpoints!), so here's a pull request adding the `_` default to the Responsive Styles docs.

The example isn't ✨perfect✨  (since a Box would usually have an implicit 100% width anyway) but I hope it's illustrative enough.

### DIFF
--- a/docs/responsive-styles.md
+++ b/docs/responsive-styles.md
@@ -59,7 +59,7 @@ Using Styled System with a CSS-in-JS library will generate something like the fo
 
 ## Using objects
 
-Alternatively, you can define breakpoints with aliases and use plain objects as values. Use the alias `_` to define the base value.
+Alternatively, you can define breakpoints with aliases and use plain objects as values. Any undefined alias key will define the base, non-responsive value.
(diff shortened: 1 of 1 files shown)


docs/responsive-styles.md
```diff
@@ -59,7 +59,7 @@ Using Styled System with a CSS-in-JS library will generate something like the fo
 
 ## Using objects
 
-Alternatively, you can define breakpoints with aliases and use plain objects as values.
+Alternatively, you can define breakpoints with aliases and use plain objects as values. Use the alias `_` to define the base value.
```
jxnblk: Thanks for this. In light of this API changing in the next version, and since *any* undefined key works in the current version, could you change the `_` key to `0`

grrowl: @jxnblk done! :)

jxnblk:
Thanks!

Issue #465, opened 2019-04-23
Incorrect sort in IE Edge for responsive objects


jamesmosier:
When using the responsive object for a style property, the order is incorrect in IE Edge.

For example, this: `<Flex flexWrap={{ mobile: 'wrap', md: 'initial' }}>` would output the `md` media query first in the CSS and the `mobile` media query second, therefore making it so the `mobile` media query rule **always** applies. 

I created a [jsbin](https://jsbin.com/cimavez/edit?js,console) to quickly reproduce the issue in IE Edge (I just rewrote some of the code to simplify things). The output from running this in Edge looks like: 

<img width="536" alt="Screen Shot 2019-04-23 at 11 13 49 AM" src="https://user-images.githubusercontent.com/2854919/56592571-f2003b00-65b8-11e9-894c-612bc91b67d2.png">

While running this in Chrome outputs:

![Screen Shot 2019-04-23 at 11 15 09 AM](https://user-images.githubusercontent.com/2854919/56592638-13f9bd80-65b9-11e9-994c-4d3018356935.png)

If i reverse the order of the properties in the object (`md` first then `mobile`) it works as expected. 

Commenting out the [`styles.sort()`](https://github.com/styled-system/styled-system/blob/master/src/index.js#L83) fixes it for both Chrome and Edge, but I wanted to get any thoughts before proceeding.

stinodes:
Hmm. changing the order of the responsive styles seems to break it on other browsers (at least it seems to in my case).  
Have you figured anything else out to get this to work as expected somewhat painlessly?

jamesmosier:
Unfortunately not @stinodes ... I ended up just writing regular styled-components for anything responsive.

jxnblk:
I believe a fix for this has been published, and the changes in #513 should make this a non-issue. Feel free to reopen this or another issue if similar problems are still happening

PR #495, opened 2019-05-20
#465 Avoid style sort in IE/Edge

stinodes:
#465 demonstrates a bug with responsive object styles, where the generated breakpoints would be ordered incorrectly in IE/Edge.  
This behavior is fixed by not sorting the styles when using either of these browsers.  

I've added the `detect-browser` dependency to detect the current browser (by name), and added a conditional around the `sort`-call.  
This has fixed the issues for my project.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,6 +2,7 @@ import PropTypes from 'prop-types'
 
 const userAgent = global.window ? global.window.navigator.userAgent : ''
 const isIE = /(MSIE )|(Trident\/)|(Edge\/)/.test(userAgent)
+const compareForIE = (a, b) => (isIE ? -1 : 1)
 
 export const defaultBreakpoints = [40, 52, 64].map(n => n + 'em')
 
@@ -107,7 +108,7 @@ export const style = ({
           styles.push({ [media]: rule })
         }
       }
-      if (!isIE) styles.sort()
+      styles.sort(compareForIE)
     }
 
     return mergeAll(...styles)
--- a/src/index.js
+++ b/src/index.js
@@ -1,8 +1,8 @@
 import { detect } from 'detect-browser'
 import PropTypes from 'prop-types'
 
-const browser = detect() || {}
-const isIE = browser.name === 'edge' || browser.name === 'ie'
+const browser = detect()
+const isIE = browser.name && (browser.name === 'edge' || browser.name === 'ie')
 
 export const defaultBreakpoints = [40, 52, 64].map(n => n + 'em')
 
@@ -108,7 +108,7 @@ export const style = ({
           styles.push({ [media]: rule })
         }
       }
-      !isIE && styles.sort()
+      if (!isIE) styles.sort()
     }
 
     return mergeAll(...styles)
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -104,7 +108,7 @@ export const style = ({
           styles.push({ [media]: rule })
         }
       }
-      styles.sort()
+      if (!isIE) styles.sort()
```
jxnblk: Thanks! I think it'd be better to not add an additional dependency and do browser checks for this sort of thing. I assume there's a way to handle this sort in a way that should work cross-browser

stinodes:
Thanks for taking a look! 
I assumed that'd be the case. I can take a look tomorrow how to handle the detection part without adding the dependency, and update the PR.

jxnblk:
IIRC, it'll probably require a compare function argument that returns a number or boolean – can't remember which one it is...

stinodes:
I have removed the detect-browser library, and have opted for a regex to check for the names of either IE-identifiers.  
Might this be worth writing a test for (basically testing the regex used against some example strings)?  

Regardless, let me know if this is fine, or you'd rather have an implementation using `indexOf`!

jxnblk:
Can you try using a comparison function instead of browser detection? i.e. I *think* this should work cross-browser

```js
const compare = (a, b) => {
  if (a < b) return -1
  if (a > b) return 1
  return 0
}
// styles.sort(compare)
```

stinodes:
Oh, like that. I didn't quite what you meant, that you meant for it to be passed to `sort`.

src/index.js
```diff
@@ -1,5 +1,9 @@
 import PropTypes from 'prop-types'
 
+const userAgent = global.window ? global.window.navigator.userAgent : ''
+const isIE = /(MSIE )|(Trident\/)|(Edge\/)/.test(userAgent)
+const compareForIE = (a, b) => (isIE ? -1 : 1)
```
jxnblk:
```suggestion
const compare = (a, b) => {
  if (a < b) return -1
  if (a > b) return 1
  return 0
}
```

src/index.js
```diff
@@ -104,7 +108,7 @@ export const style = ({
           styles.push({ [media]: rule })
         }
       }
-      styles.sort()
+      styles.sort(compareForIE)
```
jxnblk:
```suggestion
      styles.sort(compare)
```

src/index.js
```diff
@@ -1,5 +1,9 @@
 import PropTypes from 'prop-types'
 
+const userAgent = global.window ? global.window.navigator.userAgent : ''
```
jxnblk:
```suggestion
```

src/index.js
```diff
@@ -1,5 +1,9 @@
...
+const isIE = /(MSIE )|(Trident\/)|(Edge\/)/.test(userAgent)
```
jxnblk:
```suggestion
```

stinodes:
> I _think_ this should work and not require any browser checks

Yep, you are right. I made the change and tried it out and it seems to work as you predicted!  
Way cleaner this way, and sort of frustrating it was such a small thing. Thanks!

jxnblk:
Awesome! Sorry if my original comments weren't clear enough, and thanks again!

stinodes:
Got there eventually, it's all good!
Thanks for merging!

PR #513, opened 2019-05-27
V5

### LINKED ISSUE #444: Performances

### LINKED ISSUE #475: v5 Roadmap
This is a WIP issue to track potential changes for the next major version.

- [x] Remove `propTypes` from core 
- [x] Add optional package for React `propTypes`
- [x] Remove conversion of unitless numbers into pixels (most CSS-in-JS libraries support this by default now)
- [x] Setup monorepo with Yarn workspaces and Lerna
- [x] Split style functions into separate packages #281
- [x] Create separate `variants` package
- [x] Create separate `propTypes` package
- [x] Create utilities package with `themeGet`, etc
- [x] ~~Remove support for implicit breakpoints~~
- [x] Use `sizes` key per the [Theme Specification](https://system-ui.com/theme/)
- [x] ~~(Under consideration) remove fractional width support – if you make use of this feature, please speak up~~
- [x] Move experimental `css` and/or JSX pragma packages into monorepo
- [x] Move `should-forward-prop` package into monorepo

**Notes**

- For the monorepo setup, and all-in-one package should be created to make migration from v4 to v5 as easy as possible

cc @neoziro

jxnblk:
- Improved performance of core module
- Splits core module and style functions into separate packages for smaller bundling options
- Uses microbundle for main `styled-system` package
- New custom styles API
- Mostly backwards compatible with v4
- Uses `theme.sizes` scale for `width`, `height`, `minWidth`, `minHeight`, `maxWidth`, `maxHeight`, and `size` props

**Breaking**

- Removes `propTypes` from core (an optional package will be provided)
- Moves `themeGet` to a separate package `@styled-system/theme-get`
- No longer converts numbers to strings with `px` unit (most CSS-in-JS libraries now handle this)
- Removes these utilities: `isObject`, `is`, `px`, `num`, `createMediaQuery`, `defaultBreakpoints`, `merge`, `cloneFunction`, `mapProps`
- No longer returns `null` for empty values (now returns `{}`)
- Replaces width and height theme scales with `theme.sizes`

**To do**

- [x] Update docs #521
- [x] Update should-forward-prop implementation #522 
- [x] Add `prop-types` package
- [x] Split up style packages

Fixes #475 
Fixes #444


cereallarceny:
Hell yeah.

PR #516, opened 2019-05-29
Add cursor to list of props to not forward

kevinSuttle:
Is this valuable?

### DIFF
--- a/packages/styled-system/props.js
+++ b/packages/styled-system/props.js
@@ -89,6 +89,7 @@ module.exports = [
   'right',
   'bottom',
   'left',
+  'cursor',
   'buttons',
   'textStyles',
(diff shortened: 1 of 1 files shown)


jxnblk:
Thanks, but `cursor` isn't a styled-system prop, and I don't think this library should prevent that from being forwarded. There could be a component that intentionally accepts a `cursor` prop

kevinSuttle:
Oh right, my mistake. I thought I had already added it to Box as a custom prop. 0/2 today. 🤦‍♂

jxnblk:
No worries. FWIW, if you're adding custom props, you can wrap the `shouldForwardProp` function and filter out any custom props you've created

PR #518, opened 2019-05-31
Use default breakpoints variable

sebald:
Super minor detail: Getting the breakpoints didn't use the `defaultBreakpoints` from the start of the file, but rather had them declared inline.


jxnblk:
Thanks!

PR #520, opened 2019-05-31
Create new `shadow` package

mfix22:


jxnblk:
Yeah, I'm sort of unsure what to do about some of the miscellaneous properties in the library, there's also `opacity` and `overflow` which seem like they don't clearly fit anywhere...

jxnblk:
Looking at https://primer.style/components/docs/system-props – maybe `overflow` belongs in the `layout` package...

mfix22:
If you don't think these props should even be included, let me know. I don't think I have used any of the props specifically. 

If I were to take a stab at putting them _somewhere_ I would say:
* `overflow`: `layout`
* `opacity`: `color`

But it doesn't feel like a 100% fit.

Another random thought: what if we put `boxShadow` in `borders`? Especially with things like box-shadow `inset`, lots of people use it to replace border effects. Thoughts?

jxnblk:
Yeah, I think it'd make sense to have a `shadow` package like you have here and moving `overflow` to `layout` and `opacity` to `color`

packages/shadow/src/index.js
```diff
@@ -0,0 +1,16 @@
...
+    scale: 'shadows',
+  },
+})
+
+export const textShadow = system({
+  textShadow: {
+    property: 'textShadow',
+    scale: 'shadows',
+  },
+})
```
jxnblk: Thanks for working on this! For the v5 packages, I've been grouping all the utilities into categories so that one import/function will add several style props. E.g. with `typography`, it adds `fontFamily`, `fontSize`, `lineHeight`, etc. as props. I think this `shadow` package could follow the same format, where one `shadow` function adds both `boxShadow` and `textShadow` props. To do the same here, it'd mean combining the two `system` calls into a single one and merging the two configuration options

jxnblk: Going to merge this as-is and combine these in the v5 branch

PR #528, opened 2019-06-03
Fix system export

liamcmitchell-sc:
[Documentation](https://styled-system.com/api#system) says that `system` should be exported from `styled-system` but it's not.
It's imported but looks like it's never used so I moved it from import to export.


jxnblk:
Awesome thanks!

PR #538, opened 2019-06-04
Transform width to % value in @styled-system/css

sebald:
Add a trnsformer for `width` so it works like `@styled-system/layout`. Meaning,

- use `theme.sizes` to interpolate values
- transform values to percentages

### DIFF
--- a/packages/css/src/index.js
+++ b/packages/css/src/index.js
@@ -7,6 +7,8 @@ export const get = (obj, key = '', def, p, undef) => {
   return obj === undef ? def : obj
 }
 
+const isNumber = n => typeof n === 'number' && !isNaN(n)
+
 const defaultBreakpoints = [40, 52, 64].map(n => n + 'em')
 
 const defaultTheme = {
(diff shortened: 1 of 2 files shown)


jxnblk:
For now, this is an *intentional* omission to the `css` package's API. The thought being that it's inconsistent with the rest of the API and less likely to be familiar to people who haven't used Styled System before. That said, I'll leave this open for others to comment on

sebald:
I see. That sounds very reasonable. I was a bit confused when using `width` the first time with `@styled-system/css` because the documentation (even though its from `styled-system`) has example for %-based responsive values 🙂

But, not supporting % values makes the API, easier to follow, because there is no difference in accessing `sizes` OR transforming it to a % value when no `sizes` are specified.

Maybe the documentation should point out that the `css` package differs from the main package and point out some differences? 🤔

jxnblk:
Yeah, maybe having a *caveats* section would be good for now – the css packages docs are all in the readme if you want to take a stab at that

sebald:
Will do! No promise on the time though, but soon™️.

dapetrov:
So you don't plan to merge this for fractional width syntax working, right?

jxnblk:
Thanks again for putting this together! I think for now we're going to close this in order to keep the css package's API a little simpler

PR #539, opened 2019-06-04
Fix API Signature in Migrating Guide

peduarte:
Some minor yet significant API typo fixes.

The `system` function uses a `transition` key, where it should be `transform` ([src](https://github.com/styled-system/styled-system/blob/master/packages/core/src/index.js#L112))

### DIFF
--- a/docs/guides/migrating.md
+++ b/docs/guides/migrating.md
@@ -121,7 +121,7 @@ While the `style` utility from v4 should continue to work as expected, you can t
 
 ```js
 // v4
-const transition = style({
+const transition = styled({
   prop: 'transition',
 })
 const Box = styled('div')(transition)
@@ -131,7 +131,7 @@ const Box = styled('div')(transition)
 // v5
 const Box = styled('div')(
   system({
-    transition: true,
+    transform: true,
   })
 )
 ```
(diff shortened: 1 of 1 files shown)


docs/guides/migrating.md
```diff
@@ -121,7 +121,7 @@ While the `style` utility from v4 should continue to work as expected, you can t
 
 ```js
 // v4
-const transition = style({
+const transition = styled({
```
jxnblk: This is the v4 `style` API, so I think what's currently shown is correct

docs/guides/migrating.md
```diff
@@ -131,7 +131,7 @@ const Box = styled('div')(transition)
 // v5
 const Box = styled('div')(
   system({
-    transition: true,
+    transform: true,
```
jxnblk: Here `transition` is correct – it's used for both the component prop name *and* the CSS `transition` property with the shortcut syntax here

docs/guides/migrating.md
peduarte: You’re right, my bad

docs/guides/migrating.md
peduarte: Ohhh I see. Wow maybe it’s just me but I got really confused there. Thanks for the clarification.

jxnblk: It might not be the clearest example, so maybe it's worth using a different CSS property 🤔

peduarte: yeah, something like `whiteSpace` or `flexGrow`, I don't know...

PR #542, opened 2019-06-06
Add more docs to prop-types package

sapegin:
* Available Type Categories
* Custom Props

### DIFF
--- a/packages/prop-types/README.md
+++ b/packages/prop-types/README.md
@@ -38,7 +38,7 @@ See props of each category in [the reference table](https://styled-system.com/ta
 
 ```js
 import styled from 'styled-components'
-import { space, style } from 'styled-system'
+import { space, system } from 'styled-system'
 import propTypes, { propType } from '@styled-system/prop-types'
 
 const gridGap = system({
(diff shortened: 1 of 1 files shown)


packages/prop-types/README.md
```diff
@@ -19,3 +19,39 @@ Box.propTypes = {
...
+import styled from 'styled-components'
+import { space, style } from 'styled-system'
+import propTypes, { propType } from '@styled-system/prop-types'
+
+const gap = style({
+	prop: 'gap',
+	cssProperty: 'gridGap',
+	key: 'space',
+	scale: DEFAULT_SCALE,
+});
```
jxnblk: Thanks! This looks good, but I would like to promote the new `system` API instead of the v4 `styled` API for examples like this. E.g.

```js
const gridGap = system({
  gap: {
    property: 'gridGap',
    scale: 'space',
    defaultScale: DEFAULT_SCALE,
  }
})
const Stack = styled('div')(
  compose(space, gridGap)
)
```

sapegin:
Done!

PR #545, opened 2019-06-07
Add unit test for overriding margin/padding props

mfix22:
I thought it would be useful to test for this to ensure no regression occurs.


jxnblk:
Thank you! 🙏

PR #554, opened 2019-06-11
@styled-system/should-forward-prop: Add shadow to the list of composed functions

ericberens:
I noticed that the default `shouldForwardProp` does not exclude from the DOM props for `shadow` and variants (`buttonStyle`, `colorStyle`, & `textStyle`). This PR adds those to the list of composed functions to exclude.


jxnblk:
Thanks!

PR #556, opened 2019-06-11
WIP Add raw value argument to `@styled-system/css`

jxnblk:
This allows the second argument of `css` to accept raw style values that do not pick up values from the theme

### DIFF
--- a/packages/css/test/index.js
+++ b/packages/css/test/index.js
@@ -226,3 +226,16 @@ test('handles negative margins from scale', () => {
     marginRight: -32,
   })
 })
+
+test('raw values can be passed as second argument', () => {
+  const result = css({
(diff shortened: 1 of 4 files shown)


packages/css/test/index.js
```diff
@@ -226,3 +226,16 @@ test('handles negative margins from scale', () => {
     marginRight: -32,
   })
 })
+
+test('raw values can be passed as second argument', () => {
+  const result = css({
```
kof: this can be confusing because emotion allows to pass multiple style objects too and they are not treated differently

jxnblk: Yeah, good point. Just wanted to get a feel for this API – probably not the best approach

PR #564, opened 2019-06-18
Update space variables to include px

TimonVS:
From the [5.0 changelog](https://github.com/styled-system/styled-system/blob/master/CHANGELOG.md#v500-2019-06-02)

> No longer converts numbers to `px` strings

I think that means that all theme variables have to be updated to include a unit, at least in our project space didn't work without converting them to include `px` :)
Should all theme variables be updated to include a unit though?

### DIFF
--- a/docs/theme-specification.md
+++ b/docs/theme-specification.md
@@ -77,15 +77,15 @@ When defining space scales as an array, it is conventional to use the value `0` 
 
 ```js
 // example space scale
-space: [0, 4, 8, 16, 32, 64]
+space: ['0px', '4px', '8px', '16px', '32px', '64px']
 ```
 
 ```js
 // example space scale object
 space: {
-  small: 4,
-  medium: 8,
-  large: 16,
+  small: '4px',
+  medium: '8px',
+  large: '16px',
 }
 ```
 


TimonVS:
Two other things that have to be updated I think:

- https://github.com/styled-system/styled-system/blame/master/docs/api.md#L15
- https://github.com/styled-system/styled-system/blame/master/docs/api.md#L138

Both of those don't work anymore right? `width={1200}` doesn't get transformed to `px` for me anyways.

jxnblk:
Thanks! Most CSS-in-JS libraries like Emotion and Styled Components now convert numbers to pixel values, which is why it was removed from this library. You can see how numbers work in the docs for Styled System, but if you're using one of those two libraries with Styled System v5, I'd recommend updating the other libraries as well

TimonVS:
Thanks, I didn't know that :). We're still using Styled Components v3, I'd love to upgrade to 4 but there are some hard breaking changes (`innerRef` -> `ref`) that keep us from upgrading soon. I don't know how many other projects are still on SC3, but I know that Spectrum was still on v3 for a long time even though v4 was released, Primer components also still uses v3. Would it make sense to add a warning to the docs about newer CSS-in-JS libraries automatically converting the values to px but for older versions you'll have to include the unit?

jxnblk:
Yeah, the docs and migration guide could probably call this out more explicitly. In your case, I'd recommend staying on Styled System v4 until you're ready to upgrade Styled Components

Side note: it looked like Primer Components is using SC v4 last I checked

TimonVS:
Okay thanks for your insights! I'll stay on Styled System v4 for now :)

PR #568, opened 2019-06-19
Use system scale on negative positional properties

VinSpee:
`top`, `bottom`, `left`, & `right` now all use system values for
negative numbers.

Additionally, this change prompted a refactor of the
transforms object into a reducer to avoid all the repetition. LMK if that's not cool.


VinSpee:
Solves https://github.com/system-ui/theme-ui/issues/96

jxnblk:
Thanks! This looks great!

PR #572, opened 2019-06-19
css: check for null values

### LINKED ISSUE #571: skipping breakpoints
Hi again!

What's the best way "skip" a breakpoint? for example, I have some components which look like this:

```jsx
  <Flex
    css={{
      justifyContent: [undefined, undefined, undefined, 'space-between'],
      flexDirection: ['column', undefined, undefined, 'row'],
      fontSize: [2, undefined, 1],
    }}
  >
    …
  </Flex>
```

Is there a better way to _skip_ a breakpoint besides passing `undefined`?

jxnblk:
Fixes #571

### DIFF
--- a/packages/css/src/index.js
+++ b/packages/css/src/index.js
@@ -1,6 +1,6 @@
 // based on https://github.com/developit/dlv
-export const get = (obj, key = '', def, p, undef) => {
-  key = key.split ? key.split('.') : [key]
+export const get = (obj, key, def, p, undef) => {
+  key = key && key.split ? key.split('.') : [key]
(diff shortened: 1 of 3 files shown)


packages/css/src/index.js
```diff
@@ -129,12 +129,14 @@ export const responsive = styles => theme => {
 
   for (const key in styles) {
     const value = styles[key]
+    if (value === null || value === undefined) continue
```
kof: also the same as `if (value == null)` but you probably know that, I prefer this in libs

PR #577, opened 2019-06-24
docs(testimonial): add Miha Sedej testimonial

tresko:
Add Miha Sedej testimonial.


jxnblk:
Haha! I was only half joking, but I'll take it, thanks!

PR #589, opened 2019-06-28
Position: Use space scale for top/right/bottom/left

colebemis:
Hi, @jxnblk 👋 

This pull request connects the `top`, `right`, `bottom` and `left` style functions in the `position` category to the `space` theme scale. If there's a reason that they shouldn't use the `space` scale feel free to close this PR 😄  

I noticed this issue when I was using the `Absolute` component from `@primer/components` and expected `<Absolute top={3}>` to render a box `16px` from the top not `3px` from the top.

### DIFF
--- a/packages/position/src/index.js
+++ b/packages/position/src/index.js
@@ -1,15 +1,35 @@
 import { system } from '@styled-system/core'
 
+const defaults = {
+  space: [0, 4, 8, 16, 32, 64, 128, 256, 512],
+}
+
 const config = {
   position: true,
   zIndex: {
     property: 'zIndex',
     scale: 'zIndices',
   },
-  top: true,
-  right: true,
-  bottom: true,
-  left: true,
+  top: {
+    property: 'top',
+    scale: 'space',
+    defaultScale: defaults.space,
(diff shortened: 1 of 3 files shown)


colebemis:
Oops, saw that you already addressed this here: https://github.com/styled-system/styled-system/issues/420

Using margin or padding instead works for my use case 👍

jxnblk:
We've added support for position properties using the `space` scale in `@styled-system/css` so it could make sense to go ahead and add it to the `@styled-system/position` package as well

packages/position/src/index.js
```diff
@@ -1,15 +1,35 @@
...
     scale: 'zIndices',
   },
-  top: true,
-  right: true,
-  bottom: true,
-  left: true,
+  top: {
+    property: 'top',
+    scale: 'space',
+    defaultScale: defaults.space,
```
jxnblk: One difference to the `@styled-system/css` package is that it also allows negative values (like margin properties) for these. I don't think you need to address that here since the `getMargin` utility isn't exported, but something to think about

Issue #499, opened 2019-05-23
Overriding Default Values in Styled-System Variants 


moshemo:
NOTE:  I am referencing the following functionality in styled-system:  [https://styled-system.com/api#variants](https://styled-system.com/api#variants)

I have the following buttons variant in my `theme.js` file:

```
...
  buttons: {
    flat: {
      color: '#343434',
      boxShadow: 'none',
      backgroundColor: 'transparent',
      transition: 'background-color .2s',
     },
  }
```

This gives a nice, default styling for a flat button.  However, there will be times that I want to override the default color.  Say, something like this:

`<Button variant="flat" color="primary">Flat Button</Button>`

However, this does not work -- because (if I understand correctly), one cannot use props with variants (is that correct?).  If so, I am wondering if there is a way override the styling for `variant="flat"` on a case-by-case basis?

jxnblk:
You can add an additional style function to support style props with a variant prop. e.g. `styled('div')(buttonStyle, color)` – that way the end component supports both the `variant` and `color` props

moshemo:
I think I'm doing that -- but it still doesn't work.  

Here is the relevant sections of my `Button.js` file:

```
import styled from 'styled-components'
import {
  ...  
  buttonStyle,
  color,
  ...
} from 'styled-system'

import { buttonDefaults, buttonSize } from '../styles/themes/buttons'

const Button = styled('button')`
   ...
  ${buttonSize}
  ${color}
   ...
`

Button.propTypes = {
   ...
   ...color,
  ...buttonStyle,
}

Button.defaultProps = {
  ...buttonDefaults,
  size: `medium`,
}

export default Button

```

And here is the relevant sections of my `buttons.js` file:

```
import { variant } from 'styled-system'

import colors from './colors'

export const buttonDefaults = {
  bg: colors.dark,
  ...
  color: colors.light,
  ...
}

export const buttonStyles = {
  flat: {
    color: colors.dark,
    boxShadow: 'none',
    backgroundColor: 'transparent',
    transition: 'background-color .2s',
  },
  ...  
}
...
```
Or, if you want, you can see the actual code here:
[https://codesandbox.io/s/learningstyledsystemandrebass-8j6im?fontsize=14](https://codesandbox.io/s/learningstyledsystemandrebass-8j6im?fontsize=14)

Either way, I can't figure out why it doesn't work.

jxnblk:
[...]

PR #590, opened 2019-07-01
test(styled-system): default props order affects styles

angeloashmore:
Failing test related to #499.

The snapshot test has been manually altered to move `padding: 0` to the top of the list in the first test.

Note the order of the props in the tests.

## Problem

When setting default props on a component using `styled-system`, the default props may override styles defined using the style props.

```jsx
import styled from '@emotion/styled'
import { space } from 'styled-system'

const Comp = styled('div')(space)
Comp.defaultProps = { p: 0 }

// Results in Comp with no padding as if p={0} were applied after px and py.
<Comp px={2} py={1} />

// Results in Comp with padding.
<Comp p={0} px={2} py={1} />
```

If `Comp` is used with props from space, such as `px` and `py`, the generated styles will be overridden by `defaultProps.p` since `padding` takes precedence over `padding-left` and the other more-specific properties.

## Why this happens

React applies default props to the end of the props list if it has not already been defined.

```jsx
Comp.defaultProps = { p: 0 }

// Example A: Relies on defaultProps.
<Comp px={2} py={1} />
// props = { px: 2, py: 1, p: 0 }

// Example B: Does not rely on defaultProps.
<Comp p={0} px={2} py={1} />
// props = { p: 0, px: 2, py: 1 }
```

`v5` iterates over props in the order they are defined.

In Example A above, the following styles are generated:

```css
.exampleA {
  padding-left: 8px;
  padding-right: 8px;
  padding-top: 4px;
  padding-bottom: 4px;
  padding: 0;
}
```

Where we would expect it to provide `padding: 0` at the beginning of the styles instead.

```css
.exampleB {
  padding: 0;
  padding-left: 8px;
  padding-right: 8px;
  padding-top: 4px;
  padding-bottom: 4px;
}
```

## Possible Solutions

- Iterate over configs instead of props. Slower performance and relies on order of configs rather than props (i.e. less flexible).
- Iterate over props in reverse order. Could produce unexpected style ordering.
- Recommend not using default props. Instead, compose components with default props already applied but overridable. E.g. `const Comp = props => <CompBase {...CompBase.defaultProps} {...props} />`


jxnblk:
Sorry for the delay and thanks for looking into this! Not sure why this "failing" test is passing on CI. Can you confirm that it's failing locally?

angeloashmore:
It looks like the CI checks are just for building the site on Netlify. Is there a special commit message that triggers the Jest tests on CI?

angeloashmore:
Confirmed tests are failing on the PR:

```
FAIL packages/styled-system/test/defaultProps.js
  ● application of default props should not affect order of styles › default props not provided to instance

    expect(received).toMatchSnapshot()

    Snapshot name: `application of default props should not affect order of styles default props not provided to instance 1`

    - Snapshot
    + Received

    @@ -1,11 +1,11 @@
      .emotion-0 {
    -   padding: 0;
        padding-left: 8px;
        padding-right: 8px;
        padding-top: 4px;
        padding-bottom: 4px;
    +   padding: 0;
      }

      <div
        className="emotion-0"
        p={0}

      15 |     const props = Object.entries(tree.toJSON().props)
      16 |
    > 17 |     expect(tree).toMatchSnapshot()
         |                  ^
      18 |     expect(props).toEqual([
      19 |       ['px', 2],
      20 |       ['py', 1],

      at Object.toMatchSnapshot (packages/styled-system/test/defaultProps.js:17:18)

 › 1 snapshot failed.
```

jxnblk:
@angeloashmore oh crap! that's my bad – looks like the tests didn't trigger at all on this PR. Let me look into it

jxnblk:
Thanks for this, moved your commit to a new PR in #638 to continue looking into the issue!

codeincontext:
It looks like https://github.com/styled-system/styled-system/pull/638 doesn't validate the order of the properties and this problem still exists?

Seems to be blocking us from updating to v5

Issue #558, opened 2019-06-14
@styled-system/css with breakpoint objects


ruaridhnewman:
I am having an issue when using `@styled-system/css` where using breakpoint objects in our theme causes `css` to error with `breakpoints.map is not a function`. Our theme has something like this in it:

```
const theme = {
    breakpoints: {
        s: '1px',
        m: '2px',
        l: '3px',
        xl: '4px'
    }
}
```

which works for everything else, but just seems to break when we use the `css` functionality as described in the docs.

Removing the breakpoint object array from the theme fixes it (as I assume it will fallback to the default breakpoints), and also changing it to be an array instead of an object fixes the error.

I guess the `css` package just needs to be updated to handle breakpoint objects as well as arrays, or do you think I am doing something incorrect?

jxnblk:
The css package does not currently handle plain objects for `theme.breakpoints`.

In style objects, you can use standard media query syntax as an alternative to the array syntax

jxnblk:
Thanks for bringing this up! After considering this a bit, I think since the `@styled-system/css` package only supports arrays for responsive values, it doesn't really make sense to define breakpoints as a plain object when using this package. I'd like to add some utilities to make working with the `@media` rule syntax a little bit more ergonomic (which will require some exploration), but I'm going to close this issue for now

frederickfogerty:
Hey @jxnblk, I understand that the `css` package is different to the main styled-system library, but this package provides the supposed solution to a real problem we have, and I was wondering how you would suggest we solve it without the use of this package.

Problem: **using styled-system props with a non-styled-system child**

```js
const UnstyledCard = ({ className } : { className?: string }) => <div className={className} />

const StyledCard = styled(UnstyledCard)(
  css({
    px: '12px'
  })
)
// we can't use .defaultProps here or anything, because the UnstyledCard doesn't support styled-system props
// I also think that adding the styled-system helpers to every component doesn't make much sense either.
```

PR #591, opened 2019-07-01
@styled-system/css with breakpoint objects theme

maximtmartin:
Using breakpoint objects in the theme causes css to error with `breakpoints.map is not a function`. #558

### DIFF
--- a/packages/css/src/index.js
+++ b/packages/css/src/index.js
@@ -125,28 +125,53 @@ const transforms = [
 export const responsive = styles => theme => {
   const next = {}
   const breakpoints = get(theme, 'breakpoints', defaultBreakpoints)
-  const mediaQueries = [
-    null,
-    ...breakpoints.map(n => `@media screen and (min-width: ${n})`),
-  ]
+  let mediaQueries
+  if(Array.isArray(breakpoints)) {
+    mediaQueries = [null, ...breakpoints.map((n) => `@media screen and (min-width: ${n})`)]
+  }
+  else {
+    mediaQueries = {}
+    for (const key in breakpoints) {
+      mediaQueries[key] = `@media screen and (min-width: ${breakpoints[key]})`
+    }
+  }
 
   for (const key in styles) {
     const value = styles[key]
     if (value == null) continue
-    if (!Array.isArray(value)) {
-      next[key] = value
-      continue
-    }
-    for (let i = 0; i < value.length; i++) {
-      const media = mediaQueries[i]
-      if (value[i] == null) continue
-      if (!media) {
-        next[key] = value[i]
+    if(typeof value === 'object') {
+      if (Array.isArray(value)) {
+        for (const i in value) {
+          const media = mediaQueries[i]
+          if (value[i] == null) continue
+          if (!media) {
+            next[key] = value[i]
+            continue
+          }
+          next[media] = next[media] || {}
+          next[media][key] = value[i]
+        }
         continue
       }
-      next[media] = next[media] || {}
-      next[media][key] = value[i]
+      else {
+        for (const k in value) {
+          const media = mediaQueries[k]
+          if (value[k] == null) continue
+          if(k === '_' && value._) {
+            next[key] = value._
+            continue
+          }
+          if (!media) {
+            next[key] = value
+            continue
+          }
+          next[media] = next[media] || {}
+          next[media][key] = value[k]
+        }
+        continue
+      }
     }
+    next[key] = value
   }
 
   return next
--- a/packages/css/test/index.js
+++ b/packages/css/test/index.js
@@ -37,6 +37,15 @@ const theme = {
   },
 }
 
+const themeWithObjectBreakpoint =  {
+  ...theme,
+  breakpoints: {
+    xs: '40em',
+    sm: '52em',
+    md: '64em',
+  }
+};
+
 test('returns a function', () => {
   const result = css()
   expect(typeof result).toBe('function')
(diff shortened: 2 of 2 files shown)


packages/css/src/index.js
```diff
@@ -125,28 +125,53 @@ const transforms = [
...
+          next[media][key] = value[i]
+        }
+        continue
+      }
+      else {
+        for (const k in value) {
+          const media = mediaQueries[k]
+          if (value[k] == null) continue
+          if(k === '_' && value._) {
+            next[key] = value._
```
peduarte: Why are you checking specifically for `_` here? 

Is this the object convention to specify the "mobile first" property? (equivalent to the first value in an array prop)

Would the following code work:
```
css({
    color: 'primary',
    fontSize: { default: 2, small: 3, medium: 4 }
})
```

packages/css/test/index.js
```diff
@@ -77,6 +86,23 @@ test('returns system props styles', () => {
   })
 })
 
+test('[theme:object breakpoints] returns system props styles', () => {
```
peduarte: What is this array-like convention in the test description?

maximtmartin: I just copied the tests and used the theme with object breakpoints. Don't really know what the best approach is to describe the test.

packages/css/src/index.js
maximtmartin: Good point!

` _` is probably not the default "mobile first" property. I'll have to check if I can rewrite that part.

The whole `@styled-system/css` feels a bit outdated tho..

peduarte: Cool, let me know when you do and I can re-review 👍

packages/css/test/index.js
peduarte: I think for simplicity sake, you can augment the previous test ("returns system props styles"), and add an extra CSS property using object styles.

jxnblk:
Thanks for looking into this! I'll try to go through and review this soon

> I feel like this package should somehow rely on the Core functions to parse the values correctly.

For context, the `css` package is a newer, completely separate implementation from `core` and doesn't really work the same way. It's intended to be a smaller/standalone package so I wouldn't want to introduce any dependencies for this functionality

packages/css/src/index.js
```diff
@@ -125,28 +125,53 @@ const transforms = [
...
+            next[key] = value[i]
+            continue
+          }
+          next[media] = next[media] || {}
+          next[media][key] = value[i]
+        }
+        continue
+      }
+      else {
+        for (const k in value) {
```
jxnblk: From an API standpoint at this time, I do *not* want to allow object based syntax for the `style` objects (which are a bit different from props) that the `css` package uses. The array syntax is optional and there is already an object-based approach using native media query syntax, e.g.

```js
css={css({
  width: '100%',
  '@media screen and (min-width: 40em)': {
    width: '50%',
  }
})}
```

I think this adds API surface area in a way that there are two ways to do the same thing, e.g.

```js
{
  width: {
    xsmall: '100%',
    small: '50%',
  },
  '@media screen and (min-width: 40em)': {
    '33%'
  }
}
```

The above approach could be tied into the `theme.breakpoints` scale with some sort of helper for the media query key, which I think would make more sense from an API perspective

packages/css/test/index.js
```diff
@@ -37,6 +37,15 @@ const theme = {
   },
 }
 
+const themeWithObjectBreakpoint =  {
+  ...theme,
+  breakpoints: {
+    xs: '40em',
+    sm: '52em',
+    md: '64em',
+  }
```
jxnblk: For clarification, I think this *should* be supported

maximtmartin:
I don't think it's possible to allow the object breakpoints into the css package, because you have to know what the default breakpoint key is.

@jxnblk I see where you want to go with this package. This is how I'm using it at the moment.

```
css({
  width: '100%',
  '@media screen and (min-width: 40em)': {
    width: '50%',
  }
})
```

Something I would like to use is a util function to convert an object with multiple css properties and breakpoint objects values into media queries.

I think this would be a good addition:
```
{
  width: {
    xsmall: '100%',
    small: '50%',
  },
  '@media screen and (min-width: 40em)': {
    '33%'
  }
}
```

I think this PR can be closed?

maximtmartin:
@jxnblk @peduarte 

At the moment I'm solving the object breakpoints problem using `createParser` from `@styled-system/core`. Example: `css(createParser({ [prop]: sx })(props))`. In this way my breakpoint object gets converted to media queries.

The only change in the `@styled-system/css` package I would need so this would work is something like this inside the `responsive` function:
```
  let mediaQueries: { [key: string]: any };

  if (Array.isArray(breakpoints)) {
    mediaQueries = [null, ...breakpoints.map(n => `@media screen and (min-width: ${n})`)];
  } else {
    mediaQueries = {};
    for (const key in breakpoints) {
      mediaQueries[key] = `@media screen and (min-width: ${breakpoints[key]})`;
    }
  }
```

The only reason I would need this is because it throws an error when you use the object breakpoint syntax in your theme. 

My question is if `@styled-system/css` should support the breakpoint array syntax?
- if not we can simply remove the `responsive` function.
- If you want to support it why don't you support the object breakpoint syntax?

If you want to support the object breakpoint syntax, then I think I have a solution for the problem I'm currently facing with the default breakpoint. You can always check if an object is a breakpoint object or something else, because you are never writing for example `fontSize: { default: 2 }` (you would write `fontSize: 2` ), but always in combination with breakpoints (`fontSize: { default: 2, md: 4 }`).

jxnblk:
Thanks again for working on this! And, sorry for the delay!

Based on me quickly reading through the original issue #558 and not thinking through all implications myself, I thought this was something we would want to support. But since the `css` package is a bit different than the core Styled System package, I don't think it should try to handle e custom keys in `theme.breakpoints`.

I don't think the `css` package should support responsive objects as values because:

1. There's a standard way to do this with `@media`
2. Responsive array values are already supported as an ergonomic shortcut
3. Mixing nested style objects with responsive objects seems like it could lead to confusing-looking code

That said, we could look into a helper utility for adding the `@media screen and (min-width: N)` keys for standard media query syntax, but I don't think it'll need to be built into the core functionality in the `css` package.

Going to close this out for now, but feel free to open a new issue if you'd like to discuss this further

PR #595, opened 2019-07-04
Enable dynamically changed breakpoints for nested themes 

### LINKED ISSUE #592: Nested themes for local breakpoints not working
I've noticed something weird with nested themes and their respective breakpoints. We have a "base" theme with the default `const breakpoints = ['40em', '52em', '64em']`. Sometimes, we want to use "local" overrides of the base theme, e.g. for a component that might have only one breakpoint which is not in the array above. We do this like this (There's one other `ThemeProvider` further up the component tree with the "default" theme):

```jsx
const Example = () => {
  const localTheme = theme => ({
    ...theme,
    breakpoints: ['11em', '22em', '33em'],
  })

  return (
    <ThemeProvider theme={localTheme}>
      {/* Should be shown above 11em */}
      <Text display={['none', 'block']}>Hola.</Text>
    <ThemeProvider>
  )
}
```

While this works completely fine for one component, it behaves really strangely in one other.

Using the above "local theme" with:
```jsx
<Text
  color={['white', 'hotpink', 'white']}
  fontSize={[1, 2, 3]}
>
  Hola.
</Text>
```

... results in:

image

So, the `color` prop uses the "local theme" values, but `fontSize` uses the default breakpoints. Am I missing something obvious here?

gregoralbrecht:
## Problem
[Starting with v5, style functions cache `theme.breakpoints`](https://github.com/styled-system/styled-system/blob/master/CHANGELOG.md#v500-2019-06-02), which
breaks nested themes with varying breakpoints.

## Possible Solution
I'm not too sure about this, so here goes a first draft:
* If you nest themes (i.e. `<ThemeProvider>`s`), and the child theme provides a new array of breakpoints, the cache should be updated.
* If you nest one more level and don't provide new values, breakpoints will fall back to the default values.

I'm not sure if this is ideal, because I don't know why the cache was implemented in the first place. If the intention was to avoid an unnecessary `get` call, then this solution won't work because it checks whether new breakpoints have been provided with `get`. I feel like I'm in over my head here, so any help is much appreciated :)

Closes #592

### DIFF
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -29,6 +29,8 @@ export const createParser = config => {
   const cache = {}
   const parse = props => {
     let styles = {}
+    const isCacheDisabled = props.theme && props.theme.disableStyledSystemCache
+
     for (const key in props) {
       if (!config[key]) continue
       const sx = config[key]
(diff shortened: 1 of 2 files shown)


packages/core/src/index.js
```diff
@@ -36,14 +36,28 @@ export const createParser = config => {
       const scale = get(props.theme, sx.scale, sx.defaults)
 
       if (typeof raw === 'object') {
-        cache.breakpoints =
-          cache.breakpoints ||
-          get(props.theme, 'breakpoints', defaults.breakpoints)
+        // If a child theme provides breakpoints, the cached ones should be updated
+        const shouldCacheBeBusted = get(props.theme, 'breakpoints', [])
```
jxnblk: Thanks for looking into this! And yeah, the whole point of the caching was performance related so this looks like it negates a lot of that. This line in particular I think could be hoisted up above the `for` loop, and I suspect the conditional could be adjusted to minimize the impact. There should be a benchmark suite in the repo if you have time to run tests with different approaches. If the caching here doesn't make a huge impact, it might be something we can get rid of altogether

gregoralbrecht: Thanks for your feedback! I finally got around to make some changes to this PR to address your comments:

1. It now checks whether there's a key named `disableStyledSystemCache` in the provided `theme`. If not, is uses the cache just like before. If it is present, it takes the breakpoints from the provided theme (with a fallback to the defaults)
2. Concerning the benchmarks – I ran those with different implementations and couldn't really tell a significant difference in most cases. However, my results seemed to be a bit random anyway. The numbers differed quite a bit, on the same machine and without any significant background processes. If you have a way to run this more reliably, maybe we could make sure I didn't break anything performance-wise.

PR #596, opened 2019-07-05
add nonprefixed grid gap properties

chrisdhanaraj:
The latest gap spec strips the 'grid' prefix from the various gap properties (https://developer.mozilla.org/en-US/docs/Web/CSS/gap); this PR just augments the list with these properties as well


jxnblk:
Awesome, thanks!

PR #607, opened 2019-07-17
Add alias for buttonStyles

siddharthkp:
> The built-in variants use the following props and theme keys:
>
> | Function Name | Prop        | Theme Key     |
> | ------------- | ----------- | ------------- |
> | `textStyle`   | `textStyle` | `textStyles`  |
> | `colorStyle`  | `colors`    | `colorStyles` |
> | `buttonStyle` | `variant`   | `buttons`  🤔 |

I'm curious to know why the button styles are called `buttons` instead of `buttonStyles` like the other two variants? Looks like an intentional API decision.

If it makes sense, can we add `buttonStyles` as well for consistent naming? (with backward compat)

```diff
- export const buttonStyle = variant({ key: 'buttons' })
+ export const buttonStyle = variant({ key: 'buttons' }) || variant({ key: 'buttonStyles' })
```

### DIFF
--- a/packages/variant/src/index.js
+++ b/packages/variant/src/index.js
@@ -19,6 +19,7 @@ export const variant = ({
 
 export default variant
 
-export const buttonStyle = variant({ key: 'buttons' })
+export const buttonStyle =
+  variant({ key: 'buttons' }) || variant({ key: 'buttonStyles' })
 export const textStyle = variant({ key: 'textStyles', prop: 'textStyle' })
 export const colorStyle = variant({ key: 'colorStyles', prop: 'colors' })
--- a/docs/variants.md
+++ b/docs/variants.md
@@ -69,6 +69,7 @@ The built-in variants use the following props and theme keys:
 | `colorStyle`  | `colors`    | `colorStyles`               |
 | `buttonStyle` | `variant`   | `buttons` or `buttonStyles` |
 
+
 ## Custom Variants
 
 Creating custom variants allows you to extend this API in many ways.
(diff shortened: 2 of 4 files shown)


packages/variant/src/index.js
```diff
@@ -19,6 +19,7 @@ export const variant = ({
 
 export default variant
 
-export const buttonStyle = variant({ key: 'buttons' })
+export const buttonStyle =
+  variant({ key: 'buttons' }) || variant({ key: 'buttonStyles' })
```
jxnblk: Sorry for the delay on this! It looks like Circle CI isn't running tests on some of the PRs, but this doesn't look like it would work 🤔 I might be missing something, but are the tests all passing locally?

siddharthkp: Yep, they are passing on local. I just set up tests on CI as well: https://circleci.com/gh/siddharthkp/styled-system/2

> but this doesn't look like it would work

Tell me more

jxnblk: - `variant(args)` returns a function
- `funcA || funcB` will return `funcA`
- `buttonStyle` should always return the first function (unless I'm missing something, that is)

But just noticed your tests aren't testing this function here...

siddharthkp: > variant(args) returns a function

damn, got it

jxnblk: Haha! For a second, I thought you were using something that I didn't know was possible in JS :)

jxnblk: You still interested in working on this? If not, should we go ahead and close this out?

jxnblk:
Going to close this out for now. If you wanna take a stab at renaming the API for this, feel free to reopen this PR or open a new issue ✌️

PR #583, opened 2019-06-25
WIP Use `css` utility in `variant` to allow theme values in variant

jxnblk:
Fixes #582 

- [ ] Add more unit tests to cover edge cases
- [ ] Consider overall architectural implications (currently `styled-system` does not depend on `css`)

PR #631, opened 2019-07-27
Declaring variants inline

djgrant:
Adds support for declaring variants/scales inline, rather than just in the global theme. 

This is useful when the variants are only going to be used in a single component. Declaring the variants inline helps reduce theme bloat and indirection.

Example of inline variants:

```js
const Button = styled.div`
  ${layout}
  ${variant({
    prop: 'size'
    scale: [
      { fontSize: 12, padding: 12 },
      { fontSize: 16, padding: 16 },
    ]
  })}
  ${variant({
    prop: 'type',
    variants {
      primary: { background: 'green' },
      secondary: { background: 'red'  },
    }
  })}
`
```

In hindsight, I'm not sure the introduction of another alias (`variants`) is such a good idea. I've left it in the PR in case it merits discussion.

Although this wasn't my initial objective, the implementation also provides support for writing scales inline in `system`:

```js
const myStyles = system({
  property: 'some-css-property,
  scale: [0, 1] 
});
```

If this is a feature you'd like to introduce I'd be happy to follow up with an update to the docs!

### DIFF
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -1,4 +1,5 @@
 import assign from 'object-assign'
+import css from '@styled-system/css'
 
 export const merge = (a, b) => {
   let result = assign({}, a, b)
@@ -35,7 +36,10 @@ export const createParser = config => {
       if (!config[key]) continue
       const sx = config[key]
       const raw = props[key]
-      const scale = get(props.theme, sx.scale, sx.defaults)
+      const scale =
+        typeof sx.scale === 'object'
+          ? sx.scale
+          : get(props.theme, sx.scale, sx.defaults)
 
       if (typeof raw === 'object') {
         cache.breakpoints =
--- a/packages/variant/test/index.js
+++ b/packages/variant/test/index.js
@@ -1,8 +1,4 @@
-import {
-  variant,
-  textStyle,
-  colorStyle,
-} from '../src'
+import { variant, textStyle, colorStyle } from '../src'
 import { system, compose } from '@styled-system/core'
 
 const theme = {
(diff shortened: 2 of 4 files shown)


packages/variant/test/index.js
```diff
@@ -53,6 +49,54 @@ test('variant prop can be customized', () => {
...
+    variant: 'primary',
+  })
+  const b = buttonSizes({
+    size: 0,
+  })
+  expect(a).toEqual({
+    backgroundColor: 'tomato',
+  })
+  expect(b).toEqual({
+    p: 3,
```
djgrant: If #583 is merged, this can be changed to `padding: 16`.

packages/core/src/index.js
```diff
@@ -35,7 +36,10 @@ export const createParser = config => {
       if (!config[key]) continue
       const sx = config[key]
       const raw = props[key]
-      const scale = get(props.theme, sx.scale, sx.defaults)
+      const scale =
+        typeof sx.scale === 'object'
+          ? sx.scale
+          : get(props.theme, sx.scale, sx.defaults)
```
jxnblk: Thanks for working on this! I *think* this doing the same thing as `sx.defaults` does, which picks up from the `defaultScale` option, right? I might be missing something about this change though

djgrant:
Reflecting on this, I think this problem is better solved in user land (using something like https://github.com/diegohaz/styled-tools or https://gist.github.com/djgrant/1499b23be56a1a9f37f8b653f9846ed8).

As I see it, the core idea behind Styled System is its global scales. I don't want to corrupt that; scoped component styles can be solved by other tools.

djgrant:
Sorry @jxnblk, I just saw you started a review after I closed the branch! If you feel the feature is useful I am happy to get it to a mergeable state. Your perspective on what's best for the library is going to be much better than mine!

jxnblk:
@djgrant no worries! Definitely interested in making Styled System more flexible where it makes sense. I might've initially misunderstood what this PR was adding, but I *think* this might be possible today – the docs might not reflect this use case very well, but definitely interested in exploring this a little more

djgrant:
That's very possible! I did trip up on variants when I first read the docs on them so I may not have got the complete picture. Do tag me if you come back to this or want some help/feedback on docs.

PR #632, opened 2019-07-28
Docs: Add information about padding and margin utilities 

nlopin:
Adds a description of `padding` and `margin` utilities which are exported from `@styled-system/space`.

This week I had a situation when I had to provide only margins in one of my components. I couldn't find any information in the documentation. I had to read the sources to find that `space` package also exports `padding` and `margin` subsets of the `space` utility. Later I found a mention about them in the Packages section, but it was not an obvious place for me to search.

As a result, I decided to add a small section in the Spacing section with an example and a paragraph to the Reference Table section.

Let me know what do you think.

### DIFF
--- a/docs/guides/spacing.md
+++ b/docs/guides/spacing.md
@@ -128,9 +128,32 @@ When using the component, you can adjust margin and padding in any direction nee
 
 ```jsx
 // example usage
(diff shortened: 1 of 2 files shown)


jxnblk:
Awesome, thanks!

PR #654, opened 2019-08-05
Functional values can return responsive arrays

dburles:
For https://github.com/system-ui/theme-ui/issues/229, also includes a test.


jxnblk:
Thanks!

PR #670, opened 2019-08-07
Support individual border styles

dburles:
This adds support for individual border styles, now having experience using theme-ui I feel this makes sense over functional values and border shorthand. Plus, I think it fills an unexpected gap as borderWidth, borderStyle and borderColor are already supported.

Edit: Also added support for borderTop and borderBottom left and right radius

Issue: https://github.com/system-ui/theme-ui/issues/276


jxnblk:
Thanks for this!

PR #682, opened 2019-08-09
Add overflowX and overflowY

### LINKED ISSUE #676: Support for `overflowX` and `overflowY` style props
Saw that `overflow` is supported as a style prop under [layout](https://styled-system.com/table#layout).

Was wondering if `overflowX` and `overflowY` can be added as well (can work on a PR for this if it can/should be added)

mfix22:
Closes #676 

If you don't want to support these properties, please close 🙂


jxnblk:
Thanks! Yeah, I think this makes sense as an addition to the API

Issue #630, opened 2019-07-26
"as" property removes inherited props from "styled" HOC


mikemclin:
When "extending with `styled`" as seen here (https://styled-system.com/guides/build-a-box#extending-with-styled), the `as` attribute removes all of the inherited properties from the HOC.

```js
const Box = styled.div(space);

const Card = styled(Box)({
  borderRadius: '4',
  border: '1px solid #f6f6f6',
  boxShadow: '0 2px 4px rgba(0, 0, 0, .125)',
});

<Box as="p" m="3">I have margin</Box>
<Card as="p" m="3">I do not have margin</Card>
```

* `styled-system@5.0.16`
* `styled-components@4.3.2`

This was working with styled-system 4.

jxnblk:
I suspect this is an issue with `styled-components`, but I *thought* they'd fixed this already

mikemclin:
Yeah, I would expect the issue would be with `styled-components`, since it is their function.  However, it was working with `styled-system` 4, so I thought maybe there was something behind the scenes that `styled-system` 5 was doing which was not "friendly" to the `styled` function.

Also, I just tested using `styled-components@5.0.0-beta.8`, and the issue still exists.

jxnblk:
I tried adding an integration test here: https://github.com/styled-system/styled-system/blob/master/packages/styled-system/test/styled-components.js#L40-L52

If you're able to create a PR with another test like this that fails, that could be helpful. The installed version should be `styled-components@4.3.2`

jxnblk:
This appears to work as demonstrated in this [codesandbox](https://codesandbox.io/s/styled-systembasic-demo-o5xes). Please reopen this issue with a failing reproduction if you have an example of the issue described here, if it is different

PR #683, opened 2019-08-10
Add failing test for extending intermediate styled components

mikemclin:
This adds the failing test requested in #630

### DIFF
--- a/packages/styled-system/test/styled-components.js
+++ b/packages/styled-system/test/styled-components.js
@@ -51,6 +51,21 @@ describe('styled-components', () => {
     expect(json).toHaveStyleRule('color', 'tomato')
   })
 
+  test('extended intermediate components keep their props with as prop', () => {
+    const StyledBox = styled('div')(space)
+    const Box = props => <StyledBox {...props} />
+    const Card = styled(Box)(color)
+    const json = render(
+      <Card
+        as='header'
+        m={3}
+        color='tomato'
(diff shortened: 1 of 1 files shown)


packages/styled-system/test/styled-components.js
```diff
@@ -51,6 +51,21 @@ describe('styled-components', () => {
     expect(json).toHaveStyleRule('color', 'tomato')
   })
 
+  test('extended intermediate components keep their props with as prop', () => {
+    const StyledBox = styled('div')(space)
+    const Box = props => <StyledBox {...props} />
+    const Card = styled(Box)(color)
```
jxnblk: Thanks for this! But this code does look different from the original issue. I *don't* think this sort of extension would work since there's the intermediary `Box` here and `Card` won't pass `space` props through since it's now rendering a `header`, will it? AFAIK, Styled Components can only really pass things through to other Styled Components. Also need to figure out why CI tests aren't running

mikemclin: Yes, sorry the code is different.  I didn't realize the difference when I originally created the ticket.  For what it's worth, the issue was still the same...this was working with Styled System v4, but not with v5.

jxnblk: Going to close this out since I do not think the `as` prop is intended to be passed through to components that aren't Styled Components

PR #684, opened 2019-08-12
css: Adds support for size

dburles:
Wasn't sure if it was missing intentionally, but it seems like it should be included

### DIFF
--- a/packages/css/README.md
+++ b/packages/css/README.md
@@ -90,6 +90,7 @@ Property  | Theme Key
 `height` | `sizes`
 `minHeight` | `sizes`
 `maxHeight` | `sizes`
+`size` | `sizes`
 
 ## Responsive Arrays
 
--- a/packages/css/src/index.js
+++ b/packages/css/src/index.js
@@ -32,11 +32,12 @@ const aliases = {
   py: 'paddingY',
 }
 
-const directions = {
+const multiples = {
   marginX: ['marginLeft', 'marginRight'],
   marginY: ['marginTop', 'marginBottom'],
   paddingX: ['paddingLeft', 'paddingRight'],
   paddingY: ['paddingTop', 'paddingBottom'],
+  size: ['width', 'height'],
 }
 
 const scales = {
@@ -110,6 +111,7 @@ const scales = {
   minHeight: 'sizes',
   maxHeight: 'sizes',
   flexBasis: 'sizes',
+  size: 'sizes',
   // svg
   fill: 'colors',
   stroke: 'colors',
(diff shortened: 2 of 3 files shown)


jxnblk:
Thanks! Originally, I was think that this API should stick to CSS properties only, but since the `[mp][xy]` shorthands are in there already, I think it makes sense to have `size` as well

dburles:
> Thanks! Originally, I was think that this API should stick to CSS properties only, but since the `[mp][xy]` shorthands are in there already, I think it makes sense to have `size` as well

This was my exact thought too!

dburles:
Though I do think though there is a good case that theme-ui only uses CSS properties. Especially if shorthand properties could reference theme values, then there's less need for `mpxy`, etc.

dburles:
I really think that it's worth thinking about, especially since theme-ui isn't stable yet, a breaking change is feasible.

dburles:
On a related note, this would mean that `variant` could no longer be accessed within the `sx` prop. 
However, perhaps that's not too much of a problem. One thing I've found is that it quickly breaks down once the variant needs to affect multiple elements at once.

So in aiming to define a consistent API across a component library, I opted instead to always provide a `variant` prop, that way I can handle both cases. The problem is then a source of confusion around having the two options available. I haven't yet found a nice, (ideally declarative) approach to handling these kinds of variants. What are your thoughts?

In the world of the `sx` prop only allowing CSS properties, the variant API could be built around the `variant` prop and also better accomodate the ability to cover multiple elements at once.

Here's an example of a basic idea:

```js
const variants = {
  primary: {
    wrapper: {
      bg: 'primary',
    },
    heading: {
      color: 'white',
    },
    subtitle: {
      color: 'textWhiteMediumEmphasis',
    },
  },
  secondary: {
    wrapper: {
      bg: 'none',
    },
    heading: {
      color: 'text',
    },
    subtitle: {
      color: 'textMediumEmphasis',
    },
  },
};

const CardHeader = ({
  title,
  subtitle,
  variant,
  ...props
}) => {
  return (
    <Flex
      {...props}
      sx={{
        borderTopLeftRadius: 'large',
        borderTopRightRadius: 'large',
        ...variants[variant].header,
      }}
    >
      <Box sx={{ flex: 1 }}>
        <Heading
          as="h6"
          sx={variants[variant].heading}
        >
          {title}
        </Heading>
        {subtitle && (
          <Text
            sx={variants[variant].subtitle}
          >
            {subtitle}
          </Text>
        )}
      </Box>
    </Flex>
  );
};

export default CardHeader;
```

dburles:
Hey @jxnblk, just pinging you in here in case these comments got lost in the noise 🙂

jxnblk:
Thank, I did miss this one! Mind opening a new issue with some of the thoughts here? Feel like there's some good ideas, and I don't want them to get lost

dburles:
@jxnblk cool, I'll do that, give me a couple mins

dburles:
@jxnblk https://github.com/system-ui/theme-ui/issues/294

PR #685, opened 2019-08-12
Corrects broken 'edit on github' links

dburles:
I also added the ability to hide the edit link on certain pages, as I don't think the babel-plugin repo is public?

### DIFF
--- a/docs/src/edit-link.js
+++ b/docs/src/edit-link.js
@@ -2,27 +2,51 @@ import React from 'react'
 import { Location } from '@reach/router'
 import { css } from 'theme-ui'
 
-const base = 'https://github.com/styled-system/styled-system/edit/master/docs'
+// Removes preceding and trailing slashes
+const parsePage = string => string.replace(/\/+$/, '').substr(1)
 
-const getHREF = location => {
-  if (location.pathname === '/') return base + '/getting-started.md'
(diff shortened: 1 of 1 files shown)


docs/src/edit-link.js
```diff
@@ -2,27 +2,51 @@ import React from 'react'
...
+const base = 'https://github.com/styled-system/styled-system/edit/master'
+
+// Pages that don't reference the default /docs/{page}.md
+const custom = {
+  css: '/packages/css/README.md',
+  demo: '/docs/src/pages/demo.js',
+}
+
+// Don't display the edit link on these pages:
+const ignore = ['babel-plugin']
```
jxnblk: Thanks for fixing this – it's something I forgot about! I sort of think we could remove the Babel plugin from the docs entirely (separate from this PR) since it *is* experimental and hasn't really been vetted much

jxnblk:
> I don't think the babel-plugin repo is public?

It is public, just not in the main repo since it's more of an experiment:
https://github.com/styled-system/extras/tree/master/packages/babel-plugin

docs/src/edit-link.js
dburles: Might it be worth leaving in but making a clear note that it is experimental?

PR #686, opened 2019-08-12
Add prettier:readme

dburles:
Adds `prettier:readme` to package.json and includes a commit with the result of running it.

### DIFF
--- a/packages/should-forward-prop/README.md
+++ b/packages/should-forward-prop/README.md
@@ -1,4 +1,3 @@
-
 # @styled-system/should-forward-prop
 
 Utility for filtering Styled System props with Emotion's shouldForwardProp option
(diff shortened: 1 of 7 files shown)


jxnblk:
Sorry, probably merged PRs in a bad order and caused a conflict in this one

dburles:
> Sorry, probably merged PRs in a bad order and caused a conflict in this one

Np, fixed

packages/should-forward-prop/README.md
```diff
@@ -30,20 +21,20 @@ To exclude other custom props not included in Styled System, use the `createShou
...
 
 const Box = styled('div', {
-  shouldForwardProp
-})(props => ({
-  display: props.d,
-  fontWeight: props.x ? 'bold' : null,
-}),
+  shouldForwardProp,
+})(
+  props => ({
```
jxnblk: Sorry, let me think on this PR for a bit. I feel like prettier really degrades the quality of the code examples in places like this

dburles: I see what you mean but it’s pretty minor and I’d argue that these days most people are more familiar with prettier formatted code, and the benefit of formatting across the readme’s in general would be worth it

dburles: Perhaps I can pose it another way: Would it not make sense for the examples to be formatted in the same way as they would mostly likely end up appearing when they're used? 🙂 (I am making an assumption that most people using theme-ui format their code with Prettier).

dburles: @jxnblk just pinging you on this one as it's been hanging around for a bit

PR #825, opened 2019-09-17
css: add outlineColor theme scale

damnjan:
Add ability to reference color values from theme when using outlineColor
Relevant issue: https://github.com/rebassjs/rebass/issues/735


jxnblk:
Thanks!

PR #867, opened 2019-10-04
Proposal: Add 'mxb' and 'mxt' space props

jonsherrard:
[Edit: just noticed my prettier config has changed the formatting on markdown files, will fix shortly!]

Hi there!

I'm not entirely expecting this to be merged, more of a "what do you think?" PR!

Introduces `mxb` and `mxt` space props to style margins. 

A common pattern I've seen used is the aspiration to _only_ use either `margin-top` or `margin-bottom` when spacing elements vertically.

At the moment that's a lot of:

```jsx
<Box m={2} mt={0} /> || <Box mx={2} mb={2} />
```

Which isn't that bad, but with `mxb` I can do:

```jsx
<Box mxb={2} />
```

to produce:

```css
.foo {
  margin-left: 8;
  margin-right: 8;
  margin-bottom: 8;
}
```

I very rarely need `mb` to be a different level of the scale to the `mx` value, but I suppose if people need a different level of the theme scale:

```jsx
<Box mx={2} mb={3} />
```

...then this would be kind of pointless.

Perhaps an over-optimization, but I thought I'd share to see what you think.

Thanks for `styled-system`!

### DIFF
--- a/packages/css/src/index.js
+++ b/packages/css/src/index.js
@@ -23,6 +23,8 @@ const aliases = {
   ml: 'marginLeft',
   mx: 'marginX',
   my: 'marginY',
+  mxb: 'marginXBottom',
+  mxt: 'marginXTop',
   p: 'padding',
   pt: 'paddingTop',
   pr: 'paddingRight',
@@ -35,6 +37,8 @@ const aliases = {
 const multiples = {
   marginX: ['marginLeft', 'marginRight'],
   marginY: ['marginTop', 'marginBottom'],
+  marginXBottom: ['marginLeft', 'marginRight', 'marginBottom'],
+  marginXTop: ['marginLeft', 'marginRight', 'marginBottom'],
   paddingX: ['paddingLeft', 'paddingRight'],
   paddingY: ['paddingTop', 'paddingBottom'],
   size: ['width', 'height'],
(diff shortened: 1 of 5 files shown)


packages/css/src/index.js
```diff
@@ -35,6 +37,8 @@ const aliases = {
 const multiples = {
   marginX: ['marginLeft', 'marginRight'],
   marginY: ['marginTop', 'marginBottom'],
+  marginXBottom: ['marginLeft', 'marginRight', 'marginBottom'],
+  marginXTop: ['marginLeft', 'marginRight', 'marginBottom'],
```
osdevisnot: should the third value in this array be `marginTop`?

jonsherrard: Yes. Thank you.

jonsherrard: Yes. Thank you.

jonsherrard: Yes. Thank you.

joltmode:
Why don't you go the extra mile and add `myl` and `myr`?

The-Code-Monkey:
@jonsherrard I agree with @joltmode, To make more sense we might as well add

`myl` `myr` and the same for the padding `pxb` `pxt` `pyl` `pyr`

I feel like that would make more sense to do all variations.

jonsherrard:
@The-Code-Monkey I've gone down a different route, it was more of a thinking-out-loud PR, but if you'd like to grab the code for your own work, please do!

All the best. J

PR #868, opened 2019-10-06
Experimenting with Linaria

zombispormedio:
Trying to accomplish this issue https://github.com/styled-system/styled-system/issues/640
At the beginning, I thought that it would be easy to make an example using Linaria (I hadn't used it until now), but the library doesn't have nowadays the same capabilities as Emotion or Styled Components. 
So this has finally become an experiment. 
In terms of dynamic style from props, Linaria only accepts [function interpolations](https://github.com/callstack/linaria/blob/master/docs/API.md#styled) on the value side of CSS property, and it must return the correct value. Neither accept style objects nor dynamic CSS string. 

After trying more or less to make Styled System works with Linaria, I thought that maybe this experiment could help to meditate how Styled System can be better integrated with this kind of libraries. Or perhaps it's a chance to improve Linaria in this aspect. https://github.com/callstack/linaria/issues/465

### DIFF
--- a/examples/linaria/.gitignore
+++ b/examples/linaria/.gitignore
@@ -1,0 +1,23 @@
+# See https://help.github.com/ignore-files/ for more about ignoring files.
+
+# dependencies
+/node_modules
+
+# testing
+/coverage
+
+# production
+/build
+
+# misc
+.DS_Store
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
+
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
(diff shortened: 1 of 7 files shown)


glomotion:
@zombispormedio hey there! Firstly, thanks for creating this example!

My team and I are currently looking to do something very similar to what this PR is trying to achieve, and were thinking about resurrecting this to see if it's still viable. Just wondering why you closed this out...? Was there some kind of speed bump that you slammed up against which ultimately prevented use of StyledSystem within Linaria?

Issue #706, opened 2019-08-16
Override custom variant with theme


ludovicmnji:
I'm using the utility `variant()` to define some generic styles (with styled-system v5.1.0). I would like to be able to redefine some of the variants inside my custom theme but without having to re-define all the variants. Is it possible to do that ?

This is a short example of what I want to achieve :

```js
// Heading.jsx

import styled from 'styled-components';
import { variant, color, space } from 'styled-system';

const variants = variant({
  scale: 'titles',
  variants: {
    title1: {
      fontSize: 7,
      lineHeight: 1.5,
    },
    title2: {
      fontSize: 6,
      lineHeight: 1.5,
    },
    title3: {
      fontSize: 5,
      lineHeight: 1.5,
    },
    title4: {
      fontSize: 4,
      lineHeight: 1.2,
    },
    title5: {
      fontSize: 3,
      lineHeight: 1.1,
    },
    title6: {
      fontSize: 2,
      lineHeight: 1.1,
      textTransform: 'uppercase',
    },
  }
});

const Heading = styled.h1`
  font-family: ${({ theme }) => theme.fonts.primary};
  ${variants}
  ${color}
  ${space}
`;

export default Heading;
```

```js
// theme/index.js

export default {
// ...
  titles: {
    title1: {
      fontSize: 7,
      lineHeight: 1.5,
      textTransform: 'uppercase',
    }
  },
  // ...
}
```

If I do like that, it doesn't work as I expected. I lose the rest of variants (title2, title3, title4 ...) and I have to redefine inside my theme.

jxnblk:
Currently it does not merge the inline and theme-based objects, but I could see this being a sensible addition to the API, see https://github.com/styled-system/styled-system/blob/master/packages/core/src/index.js#L48

ludovicmnji:
Thanks for the answer, I was not sure to use `variant` correctly or if it was just a bug. 

I don't know the complexity for you to implement that, but in my opinion it could be very powerfull. It could participate in writing more customizable component with an abstract way (across theme). Immediately, for my use, I'm going to write base style inside my default theme whereas component definition and rewrite each style in others themes.
  
Do you have an idea for planning this improvement ?

jxnblk:
Yeah, I don't imagine it'll be too difficult to implement, but also need to be cautious about breaking changes and performance implications (which is the more time consuming part). Going to leave this issue open for now. Thanks!

harrisrobin:
[...]

PR #886, opened 2019-10-11
Merging variant and theme-based objects

mathanpec:
This PR tries to solves for #706 . 

If a variant styles are defined in both theme object and as inline to `variant()` function, variants in the theme object will be the base and it will take the remaining variants from inline variants.

This makes it possible to override only few variants in our custom theme object and still use all the variants from inline variants and theme object without having to redefine everything in our theme object.

Please look at the test case added in this PR to understand the functionality better.

### DIFF
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -39,7 +39,7 @@ export const get = (obj, key, def, p, undef) => {
   return obj === undef ? def : obj
 }
 
-export const createParser = config => {
+export const createParser = (config, getScale = get) => {
   const cache = {}
   const parse = props => {
     let styles = {}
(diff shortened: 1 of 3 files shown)


jxnblk:
Generally this LGTM – I'd like to look a little more closely at the implementation later

mathanpec:
Hi @jxnblk, Any update on this? Would want to use this in our app.

ankeetmaini:
This is great and just what I need! Is this good to be merged @jxnblk ? :)

South-Paw:
@jxnblk any word on when you'll have time to get this in? Would be very useful.

The-Code-Monkey:
Why don't we do this using the same implementation of this that was added to theme-ui seeing as theme-ui is based off styled-system that way by adding it to this package it can be removed from the theme-ui package at the same time, rather than implementing the same functionality in two different ways.

michaelkay-wrisk:
I was wondering if there was any update on this?

mathanpec:
@jxnblk: Can you pls check this and update if there is any correction needed ?

PR #1000, opened 2019-12-05
Support theme keys containing a dot

klaasman:
If a key in the theme contains a dot (`.`) it could never be resolved because the get function woud split the string by that dot. E.g:

```js
const theme = {
  colors: {
    'my.color' : '#fff'
  }
}

// ...

<Box bg='my.color' />
```


klaasman:
I'm closing this issue due to difficulties with combining this feature with array index resolvers like `colors.blue.3`.

cheton:
@klaasman I suggest you change the color key to `blue:3` rather than `blue.3`.

PR #1010, opened 2019-12-15
ability to return an object of CSS properties in transform

Polypants:
This would allow props to apply different CSS values to multiple properties, based on the input value.

Right now the return value of `transform` is used as the CSS value for each property in `properties`. With this change, if the return value of `transform` is an object the keys and values will be applied as CSS. Allowing the assignment of different CSS values for multiple properties with one prop.


jxnblk:
Thanks! Could you open an issue to discuss this and please provide documentation, examples, and unit tests?

richardaum:
Tiny change but great! When is this being released?

PR #1023, opened 2019-12-21
Don't merge nested objects into each other

siddharthkp:
without the first empty object, it will merge properties of b into a :)

Repro: https://codesandbox.io/s/dark-haze-cremt

I saw styled-system uses lodash.merge in other places, is this utility not used anymore?


jxnblk:
Thanks! Would you mind adding a test for this to help catch if this changes again in the future?

PR #1047, opened 2019-12-28
[Docs] Fix Artsy logo

damassi:
Noticed that the Artsy logo was a broken image reference, so fixed that. Also noticed that there were no links out from the logos.


jxnblk:
Thanks! The lack of links was intentional and fairly typical for this sort of usage

damassi:
Ah, will remove (but did want to fix our broken logo image)

jxnblk:
Thanks!

PR #1115, opened 2020-02-07
Add Text Contrast function

The-Code-Monkey:
This is the contrast function to be used in conjunction with the color function.

The contrast function works by looking at the backgroundColor and setting the color prop to either `black` or `white` depending on which color is more readable, 

It cannot be used within the `compose` function as far as I am aware due to how `compose` works.

The `contrast` function will work slightly differently depending on the order you place the `color` function in, if you put `contrast` before `color` if you don't specify a text color `contrast` will set one, if you manually set one via props the color you specify will take over, this will mean if you want a component to always be black or white and not override-able then place `compose` after `color`


The-Code-Monkey:
Feel free to suggest any changes

jxnblk:
Thanks for this! Would you mind adding some context to the PR description and maybe adding a README showing how this is intended to be used?

The-Code-Monkey:
Do you want me to add the docs to the place where the rest of the api sits, here: https://styled-system.com/api or a README in the folder I added?

jxnblk:
A README in this package would be the first place to start -- just enough to demonstrate what this is for is fine

The-Code-Monkey:
I have added that readme for you. let me know if you want any changes.

The-Code-Monkey:
@jxnblk any idea as to when this will get merged, getting close as to needing it within our project.

jxnblk:
@The-Code-Monkey if you want to use this in a project, I'd recommend collocating the code in your project or publishing your own package. I haven't had time to review this PR properly, and last time I checked the README was still somewhat unclear as to the intent here

Issue #813, opened 2019-09-11
Array with `null` generates empty media query


wfifer:
Looking at docs [here](https://styled-system.com/guides/array-props/), passing `null` in an array position should not generate a media query for corresponding to that position, but that doesn't appear to be the case. I'm using latest `styled-components` and `styled-system`. It's become crucial to whittle down our generated css due to Google's low-low 50kb AMP limit, so any help would be extremely appreciated.

jxnblk:
`null` values used to not render but this causes issues – see #561 #551 #549 – without a clear way to handle this sort of issue, I'm afraid it will have to return an empty media rule for the time being

Issue #561, opened 2019-06-16
Wrong media query order


sarneeh:
Hey! First of all, thank you again for this amazing library! :smile: 

I have a weird problem. Take a look at this example:

```js
<Button m={[1, null, 3]} p={[1, 2, 3]}>
  Test1
</Button>
```

This results in:

```css
@media screen and (min-width: 40em) {
    .gtxxcJ {
        padding: 8px;
    }
}

@media screen and (min-width: 52em) {
    .gtxxcJ {
        margin: 16px;
        padding: 16px;
    }
}

.gtxxcJ {
    margin: 4px;
    padding: 4px;
}
```

Which is wrong, as the lower media query is overriding the higher one (due to order in css).

Reproduction: https://codesandbox.io/embed/styled-system-media-query-problem-qbhnd

Version: `styled-system@4.2.3`. This also happens in `5.x`, but behaves a little bit differently (you can check it by changing the version in `codesandbox`).

jxnblk:
Thanks! This *appears* to be working in Chrome (at least on my end) and inspecting the injected CSS looks like styled-components is sorting the values differently than you have above. Where did you pull that CSS block from, and could you list out relevant versions of styled-components, Emotion, etc.?

psullivan6:
I can also confirm this issue, which more precisely seems to happen when a prop using a larger breakpoint is listed before another prop using a smaller breakpoint. For instance, my similar usage:

### Similar usage
```jsx
<Box
  paddingTop={{ zero: 3, medium: 120 }}
  paddingBottom={{ zero: 220, tiny: 400, medium: 230 }}

  etc...
```

### CSS Output
_(from copy/pasting CSS code directly from the injected `style` tag in the `<head>`:)_
```css
.jYZJIX {
  ...a lot of other styles
  padding-top: 20px;
  padding-bottom: 220px;
}

@media screen and (min-width:48em) {
  .jYZJIX {
    padding-top: 120px;
    padding-bottom: 230px;
  }
}

@media screen and (min-width:20em) {
  .jYZJIX {
    padding-bottom: 400px;
  }
}
```


### My testing environment
Mac - Chrome - version `75.0.3770.142`
`styled-system` - version `5.0.5`
`styled-components` - version `4.3.2`

### Theory

[...]

Issue #551, opened 2019-06-10
Order of media queries in v5 (breakpoints object, defaultProps)


fk:
Using the latest v5, I ran into an issue of media queries being ordered wrong when used with a `breakpoints` _object_ and `defaultProps`. This worked fine in v4.

Here's a https://codesandbox.io/s/styled-system-defaultprops-breakpoints-410v5 with a reproduction. I'll hold back from explaining what is going on there for now (in the hope that the sandbox code is self explanatory).

<img width="1341" alt="foo" src="https://user-images.githubusercontent.com/21834/59212900-da563380-8bb3-11e9-9bf1-509df720c524.png">

![image](https://user-images.githubusercontent.com/21834/59213062-4769c900-8bb4-11e9-9ebd-b73452df91eb.png)

leekad:
I have the same problems, in my case:
- defaultProps now override all the props

example from `rebass`

```
const Header = styled(Text)({})

Header.defaultProps = {
  m: 0,
  mb: '8px'
}
```

So the `defaultProps` now override all `m`, `mb` when I create <Header mb={10} />

sarneeh:
I have the same issue but actually in `v4`. Unfortunately, for some reason, I can't reproduce it. It behaves like that in my project but not in any other environment :confused:

jxnblk:
I *think* this was resolved in `v5.0.18`, but feel free to reopen this if you still notice the issue:
https://codesandbox.io/s/styled-system-defaultprops-breakpoints-bs2gf

PR #1137, opened 2020-02-17
Fix media query when using css package with null responsive values

RyanTheAllmighty:
Hi,

Noticed an issue with responsive arrays when using @styled-system/css today. When null values are passed in, the order that the media queries are loaded in, can be out of order.

A detail of the issue and what is happening is in the issue https://github.com/system-ui/theme-ui/issues/577 on theme-ui project.

I looked through many issues in this repo where it was bought up, #551, #561 and #813. 

I noticed it was fixed in theme ui with PR https://github.com/system-ui/theme-ui/pull/600

This is a port of that fix, with the added tests also.

I don't have full context over the reason it may or may not have been ported from theme-ui, since it seems the css package there is identical to this one, minus this fix.

### DIFF
--- a/packages/css/src/index.js
+++ b/packages/css/src/index.js
@@ -163,12 +163,12 @@ export const responsive = styles => theme => {
     }
     for (let i = 0; i < value.slice(0, mediaQueries.length).length; i++) {
       const media = mediaQueries[i]
-      if (value[i] == null) continue
       if (!media) {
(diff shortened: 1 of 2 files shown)


jxnblk:
Thanks! The `@styled-system/css` package will be deprecated in favor of `@theme-ui/css` in the future, but I think it makes sense to get this patch in

PR #1150, opened 2020-02-23
Update `createPropTypes` from O(n^2) to O(n)

ianduvall:
I noticed `createPropTypes` was spreading the accumulator every loop. I realize this is a micro-optimization but why not. More details: https://medium.com/better-programming/the-reduce-spread-anti-pattern-fc0c1c0b23f6


ianduvall:
👋 bump

PR #1156, opened 2020-02-25
Use more human readable values for margin and padding

lolcookie:
Follows what facebook uses for react native.


The-Code-Monkey:
Nice idea give people an option as to how they want to do it.

PR #1161, opened 2020-02-28
Add missing aliases to space props

dcastil:
The props `marginX`, `marginY`, `paddingX` and `paddingY` are missing in the reference table documentation. This PR adds those props. 😊


jxnblk:
Thanks!

PR #1254, opened 2020-04-08
Styled Components now supports shouldForwardProps

rileybracken:
Adds documentation on how to implement

### DIFF
--- a/docs/guides/removing-props-from-html.md
+++ b/docs/guides/removing-props-from-html.md
@@ -31,12 +31,18 @@ const Box = styled('div', {
 
 ## Styled Components
 
-Unfortunately, Styled Components does not currently support an API to control which props are forwarded to the HTML element.
-**If you'd like to see support for this, please leave a comment on their long-running issue:**
+As of [release v5.1.0](https://styled-components.com/releases#v5.1.0) Styled Components has added a `shouldForwardProp` API. It isn't all that different from what Emotion is doing.
 
-**[Separate HTML attributes from styling props][styled components issue]**
+```js
+import styled from 'styled-components';
+import { typography } from 'styled-system';
 
-[styled components issue]: https://github.com/styled-components/styled-components/issues/439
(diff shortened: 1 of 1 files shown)


docs/guides/removing-props-from-html.md
```diff
@@ -31,12 +31,18 @@ const Box = styled('div', {
 
 ## Styled Components
 
-Unfortunately, Styled Components does not currently support an API to control which props are forwarded to the HTML element.
-**If you'd like to see support for this, please leave a comment on their long-running issue:**
+As of [release v5.1.0](https://styled-components.com/releases#v5.1.0) Styled Components has added a `shouldForwardProp` API. It isn't all that different from what Emotion is doing.
```
jxnblk, line 34: Thanks for looking into this. Since the API is identical, I think this guide should be rewritten and the two libraries do not need to be called out separately like they currently are. This not about minimum version for Styled Components is important, but the directions in this guide should be focused on using `@styled-system/should-forward-prop`

jrolfs: This makes sense, but currently the `@styled-system/should-forward-prop` package is still specific to Emotion (it depends on `@emotion/memoize` and `emotion/is-prop-valid`).

I was thinking about opening a PR to add another package with a version that works without those dependencies but I'm curious how you envision this to work moving forward.

jxnblk: Do those dependencies negatively affect the functionality when used with Styled Components?

EoleO: Sometime when using the property `as` inside a component, the function `shouldFowardProp` doesn't work well because the function `isPropValid` (from emotion) doesn't accept `as` as a valid property.

alexandernanberg: Styled components itself has as dependency on `@emotion/is-prop-valid` so it shouldn't be an issue

PR #1256, opened 2020-04-12
Remove object-assign from styled-system dependencies

znarf:
Doesn't seem to be used.


jxnblk:
Thanks! This looks like a remnant from before `@styled-system/core`

PR #1280, opened 2020-04-20
store config properties on styleFunctions

hainessss:
Instead of completely overriding the config with the style functions, keeping the original config values on the styleFunctions makes them easy to extend with simple middleware.

≈
```
  ...
  newConfig.paddingX = createStyleFunction({
        properties: newConfig.paddingX.properties,
        scale: newConfig.paddingX.scale,
        transform: (n, theme) => {
            // do stuff here.
             return newConfig.paddingX.transform(n, theme)
        }
})
```

Alternatively, exporting the original config objects makes it easy to extend as well?

### DIFF
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -152,6 +152,8 @@ export const createStyleFunction = ({
     })
     return result
   }
+  sx.properties = properties
+  sx.transform = transform
   sx.scale = scale
   sx.defaults = defaultScale
   return sx


hainessss:
@jxnblk  Does this make sense to you? Do I need to clarify anything? The question I am trying to answer with this PR is: How do I use a slightly modified version of a styled-system parser export without re-creating a whole new config for each parser or forking the repo? (namely extending the transform function to have some custom logic)

PR #1317, opened 2020-05-07
Add support for styling with pseudo selectors to core

### LINKED ISSUE #1227: Use pseudo-classing within `system`
It would be nice to be able to use `system` to produce pseudo-classed styling. There are different ways you could do this, perhaps:

```javascript
  system({
    colorHover: { property: "color", scale: "colors", pseudo: "hover" },
    bgHover: { property: "backgroundColor", scale: "colors", pseudo: "hover" },
    borderColorHover: { property: "borderColor", scale: "colors", pseudo: "hover" }
  })
```

(or perhaps using object or dot notation in the `property` properties)

becoming

```css
&:hover {
  color: <colorHover prop>;
  background: <bgHover prop>;
  borderColor: <borderColorHover prop>;
}
```

roginfarrer:
Resolves #1227

This PR adds support for applying styles with pseudo selectors. This looks a lot like [Chakra's PseudoBox API](https://chakra-ui.com/pseudobox)

```javascript
<Box
  color="red.200"
  _hover={{
    color: 'red.400'
  }}
  _disabled={{
    color: 'gray.200'
  }}
/>
```

Similar to breakpoints, styled-system will apply a set of defaults for the more common pseudo selectors. This is configurable using the `system` utility. For example, when creating a Box component:

```js
const extraProps = {
  textAlign: true,
  fontWeight: true,
}

const config = compose(
  space,
  color,
  layout,
  system(extraProps, {
    _hoverAndFocus: '&:hover, &:focus'
    _disabled: '&:disabled, &[aria-disabled="true"]',
  }
);

const Box = styled('div')(config);
```

## Todo

* [ ] Add documentation

### DIFF
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -12,24 +12,30 @@ export const merge = (a, b) => {
 }
 
 // sort object-value responsive styles
-const sort = obj => {
+const sort = (obj) => {
   const next = {}
   Object.keys(obj)
-    .sort((a, b) => a.localeCompare(b, undefined, {
-      numeric: true,
-      sensitivity: 'base',
-    }))
-    .forEach(key => {
+    .sort((a, b) =>
+      a.localeCompare(b, undefined, {
+        numeric: true,
+        sensitivity: 'base',
+      })
+    )
+    .forEach((key) => {
       next[key] = obj[key]
     })
   return next
 }
 
 const defaults = {
-  breakpoints: [40, 52, 64].map(n => n + 'em'),
+  breakpoints: [40, 52, 64].map((n) => n + 'em'),
 }
-const createMediaQuery = n => `@media screen and (min-width: ${n})`
+const createMediaQuery = (n) => `@media screen and (min-width: ${n})`
 const getValue = (n, scale) => get(scale, n, n)
+const PSEUDO_SELECTORS = {
+  _hover: '&:hover',
+  _focus: '&:focus',
+}
 
 export const get = (obj, key, def, p, undef) => {
   key = key && key.split ? key.split('.') : [key]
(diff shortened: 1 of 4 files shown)


jxnblk:
Thanks! An earlier version of Styled System had an API for pseudo selectors, but I question the usefulness of this -- how many times do you need to change a hover state dynamically based on props? I'm sure it's more than zero, but it adds some amount of API surface area (and custom syntax) to the library

roginfarrer:
Sorry for the long delay, I haven't had time to come back to this.

I think Chakra's PseudoBox is a good example of how this can be useful. It extends what the Box component (or any styled-system implementation) can do. If the API surface area is a concern, do you think there should be a fixed set of pseudo modifiers provided? Then there isn't an API, but we can then provide those features.

coach-arne:
For what it's worth I came across this pull-request when trying to implement some hover styles on cards which I wanted to be changed by props. So for me personally it would be worth the inclusion.

mariosant:
Agreed! This is extra useful!

matigda:
Hi guys, when we can expect it to be merged? :)

roginfarrer:
@coach-arne @mariosant @matigda I've forked and rewritten styled-system into a new package called [system-props](https://github.com/roginfarrer/system-props) that adds this feature as well as a few commonly requested features. Feel free to take a look—it's compatible with most of the styled-system API, and I'm still actively working on it.

matigda:
I mean I would love to know if this project is still taken care of. Seems like authors don't check it anymore.

yipei1028:
> Thanks! An earlier version of Styled System had an API for pseudo selectors, but I question the usefulness of this -- how many times do you need to change a hover state dynamically based on props? I'm sure it's more than zero, but it adds some amount of API surface area (and custom syntax) to the library

Hi @jxnblk,

I think it's worth to have this API. 🔥🔥🔥
For example, in <table> element, we might need to change each row state by props and each row present it own hover style.

I'm really stuck - do you have any idea how to solve these problem within using styled-system and styled-component?

Please take a look on this [codesandbox](https://codesandbox.io/s/demo-for-why-pseudo-selector-is-needed-s1n0h?file=/src/App.js).

Thank you for considering my request.

Yi-Pei

PR #1495, opened 2020-07-28
Design changes.

Ktoxcon:
Styled-system is a great tool for building applications using React.

There are different sources that assure that it is not favorable to use the pure black color for backgrounds or text, therefore to avoid eyestrain those who read the documentation make some changes in the colors that I hope will be taken into account..


jxnblk:
Thanks for the PR, but this really, really isn't on brand, hope that makes sense

PR #1283, opened 2020-04-25
Update styled-components's section

apperside:
Starting from version 5.1.0 styled-components now allows to filter the props

PR #1729, opened 2020-11-28
Update shouldForwardProp documentation for styled-components

### LINKED ISSUE #1615: Documentation for removing props from HTML is out of date?
On removing props from the HTML when using Styled Components the docs say this:
`Unfortunately, Styled Components does not currently support an API to control which props are forwarded to the HTML element. If you'd like to see support for this, please leave a comment on their long-running issue:`

https://styled-system.com/guides/removing-props-from-html/#removing-props-from-html-elements

However I think this is out of date. From v5.1 the shouldForwardProp was added:
https://styled-components.com/docs/api

Are the docs indeed wrong? If so Id be happy to open a PR to update if that would be helpfull?

tywmick:
Resolves #1615, and I think this fixes the concerns with PRs #1254 and  #1283, as well. Here's how the new sections read:

> ## `shouldForwardProp`
>
> Both [Emotion](https://emotion.sh/docs/styled#customizing-prop-forwarding "Emotion – Customizing prop forwarding") and [Styled Components](https://styled-components.com/docs/api#shouldforwardprop "styled-components API Reference – shouldForwardProp") include support for determining which props are forwarded to the HTML element with their `shouldForwardProp` APIs.
>
> Styled System has a optional utility that can be passed directly to either API, making sure your components do not render style props as HTML. To use it, start by installing the following package:
>
> ```sh
> npm i @styled-system/should-forward-prop
> ```
>
> ### Emotion
>
> When using Emotion, pass this utility function as an option to the `styled` HOC.
>
> ```js
> import styled from '@emotion/styled'
> import shouldForwardProp from '@styled-system/should-forward-prop'
> import { space, color } from 'styled-system'
>
> const Box = styled('div', {
>   shouldForwardProp,
> })(space, color)
> ```
>
> ### Styled Components
>
> With Styled Components, pass it as an option in the `withConfig` function.
>
> ```js
> import styled from 'styled-components'
> import shouldForwardProp from '@styled-system/should-forward-prop'
> import { space, color } from 'styled-system'
>
> const Box = styled('div').withConfig({
>   shouldForwardProp,
> })(space, color)
> ```

### DIFF
--- a/docs/guides/removing-props-from-html.md
+++ b/docs/guides/removing-props-from-html.md
@@ -6,18 +6,19 @@ For the most part, these libraries do a good job of not forwarding props that ar
 However, the Styled System API uses a few prop names that are either deprecated HTML attributes or SVG attributes and aren't intended to be rendered in the output HTML.
 While most browsers should handle these stray HTML attributes with no problem, there might be good reason to clean these up in production.
 
-## Emotion
+## `shouldForwardProp`
 
-Emotion has first-class support in its API for determining which props are forwarded to the HTML element with its [`shouldForwardProp`](https://emotion.sh/docs/styled#customizing-prop-forwarding) API.
-Styled System has a optional utility that can be passed directly to this API.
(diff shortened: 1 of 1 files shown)


luigi-derson:
I was going to create a PR for this as well. Great addition.

PR #1931, opened 2021-04-21
parseResponsiveStyle can pass the index

bichikim:


3imed-jaberi (commented):
Can you provide more information like the context and the goal of this change!

PR #2084, opened 2021-08-13
Improve css function performance

leihuang23:
This PR aims to improve the performance of the `css` function significantly. 

We've found that the `css` function is a major contributor that harms the LCP and runtime performance of our site, so every bit of performance gain from micro tuning matters.

Currently there're a few performance bottlenecks:

1. Object merging. The spread operator used to merge objects will be transpiled to `Object.assign` by Babel. This is not optimal in terms of performance. A raw `for` loop will be much more performant. Run this test to find out the difference:
```js
const assign = (target, ...sources) => {
    for (let i = 0; i < sources.length; i++) {
        const source = sources[i];
        const keys = Object.keys(source);
        for (let j = 0; j < keys.length; j++) {
            const key = keys[j];
            target[key] = source[key];
        }
    }

    return target
}

const randStr = len => Math.random().toString(36).substr(2, len)
const createObj = () => new Array(600).fill(null).reduce((obj, _, idx) => {
    obj[`${randStr(6)}-${idx}`] = randStr(10)
    return obj
}, {})

const o1 = createObj();
const o2 = createObj();

console.time('assign')
for (let i = 0; i < 1000; i++) {
    assign({}, o1, o2)
}
console.timeEnd('assign')

console.time('Object.assign')
for (let i = 0; i < 1000; i++) {
    Object.assign({}, o1, o2)
}
console.timeEnd('Object.assign')
```

2. The recursion of `css` is very inefficient. It recreates `theme`(which is expensive to calculate) and copies the same `result` in every iteration. By initiating the `theme` once and mutating the `result` in place, we can gain a huge perf improvement.

### DIFF
--- a/packages/css/src/index.js
+++ b/packages/css/src/index.js
@@ -1,20 +1,38 @@
 // based on https://github.com/developit/dlv
-export const get = (obj, key, def, p, undef) => {
-  key = key && key.split ? key.split('.') : [key]
+export var get = function get(obj, key, def, p, undef) {
+  var _key;
+
+  key = (_key = key) != null && _key.split ? key.split('.') : [key];
+
   for (p = 0; p < key.length; p++) {
-    obj = obj ? obj[key[p]] : undef
+    var _obj;
+
+    obj = ((_obj = obj) == null ? void 0 : _obj[key[p]]) || undef;
   }
-  return obj === undef ? def : obj
-}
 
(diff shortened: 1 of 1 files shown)


3imed-jaberi (approved):
For sure, this modernizes the code and improves the perf. 
LGTM +1 🚀

